PR #3, opened 2018-08-08
OPENEUROPA-1105: Implement content type event

upchuk:
This PR branched out from OPENEUROPA-1103 and it contains all its code.

Closing PR#1 which will be reviewed here.


modules/oe_content_announcement/config/install/core.entity_form_display.rdf_entity.announcement.default.yml
```diff
@@ -0,0 +1,115 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.rdf_entity.announcement.oe_ct_announcement_body
```
ademarco: This comment applies to all other similar configuration. On OpenEuropa Paragraphs we set on the convention that all configuration that is provided by our core modules should be pre-pended with`oe_` only. I'd propose the following changes:

- For content types add `oe_` prefix
- For field related configuration replace `oe_ct_` with `oe_`

The rest is ok, i.e. having bundle machine name repeated in field machine names too, I've discussed this with @brummbar and it is indeed safer this way so we make sure that we can map different RDF schemas to semantically similar fields.

modules/oe_content_announcement/config/install/core.entity_form_display.rdf_entity.announcement.default.yml
```diff
@@ -0,0 +1,115 @@
...
+    - field.field.rdf_entity.announcement.oe_ct_announcement_publish_date
+    - rdf_entity.rdfentity.announcement
+  module:
+    - field_group
+    - text
+third_party_settings:
+  field_group:
+    group_announcement:
+      children:
+        - group_announcement_content
```
ademarco: Same applies here, we should prepend machine names with `oe_`

PR #4, opened 2018-08-13
OPENEUROPA-1112: Save the provenance URI to each entity

upchuk:
### Description

* Created config + form for setting the Provenance URI for a given site
* On entity presave, RDF entities have this value saved in their new Provenance URI base field (added via alter)
* Exporting the RDF mapping for this new field on the Announcement and Event bundles
* Behat test (instead of Kernel because the latter won't work in Drupal 8.6)
* Minor scope creep: 1 step after the announcement.feature and event.feature to clear the RDF entity after being created in the step

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


oe_content.module
```diff
@@ -2,9 +2,51 @@
...
+ * Implements hook_entity_presave().
+ */
+function oe_content_entity_presave(EntityInterface $entity) {
+  // RDF entities need to be saved with a provenance URI.
+  if (!$entity instanceof Rdf) {
+    return;
+  }
+
+  $uri = \Drupal::config('oe_content.settings')->get('provenance_uri');
+  if (!$uri) {
```
brummbar: Let's throw a critical exception here. The website is unusable if the provenance uri is not set. Theoretically it should go earlier on, like when accessing the entity create/edit form, but it's ok also here.

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,145 @@
...
+  public function afterScenarioRdfCleanUp(AfterScenarioScope $scope): void {
+    // If a provenance URI was changed, restore the previous one..
+    $provenance_uri = \Drupal::state()->get('behat_rdf_test_existing_provenance');
+    \Drupal::configFactory()
+      ->getEditable('oe_content.settings')
+      ->set('provenance_uri', $provenance_uri)
+      ->save();
+    \Drupal::state()->delete('behat_rdf_test_existing_provenance');
+
+    $label = \Drupal::state()->get('behat_rdf_test_created_entity');
```
brummbar: Take a look at the [implementation in Joinup](https://github.com/ec-europa/joinup-dev/blob/develop/web/modules/custom/solution/solution.behat.inc), we can take some [stuff from there](https://github.com/ec-europa/joinup-dev/blob/develop/tests/src/Traits/RdfEntityTrait.php).

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,145 @@
...
+        $entity->delete();
+      }
+      \Drupal::state()->delete('behat_rdf_test_created_entity');
+    }
+  }
+
+  /**
+   * Configures the Provenance URI.
+   *
+   * @Given /^the site Provenance URI is set to "([^"]*)"/
```
brummbar: Let's use the new syntax for step definitions: `@Given the site provenance URI is set to :uri`.

tests/features/event.feature
```diff
@@ -26,4 +26,5 @@ Feature: Event feature
     Then I should see "http://example.com/my-event"
     Then I should see "Wed, 08/08/2018 - 12:00"
     Then I should see "Wed, 08/08/2018 - 14:00"
+    Then I delete the RDF entity with the name "My event"
```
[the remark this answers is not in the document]
ademarco: Please discard this comment, we assessed with @brummbar  and @upchuk  that this is ok.

tests/features/announcement.feature
```diff
@@ -19,4 +19,5 @@ Feature: Announcement feature
     Then I should see "My description"
     Then I should see "My body"
     Then I should see "Wed, 08/08/2018 - 01:01"
+    Then I delete the RDF entity with the name "My announcement"
```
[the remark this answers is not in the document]
ademarco: Please discard this comment, we assessed with @brummbar  and @upchuk  that this is ok.

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Creates keeps track of an RDF entity.
+   *
+   * @param array $values
+   *   Values for the RDF entity.
+   */
+  protected function createRdfEntity(array $values): void {
+    $bundle = isset($values['bundle']) ? $values['bundle'] : NULL;
+    if (!$bundle) {
+      throw new \InvalidArgumentException('No bundle provided..');
```
ademarco: Double trailing dots: `..`.

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+      throw new \Exception('The RDF entity did not get created with the proper Provenance URI.');
+    }
+  }
+
+  /**
+   * Removes a RDF entity with a given name.
+   *
+   * @Then /^I delete the RDF entity with the name "([^"]*)"$/
+   */
+  public function iDeleteTheRdfEntityWithTheName($arg1): void {
```
ademarco: Shall we use in the step `:name` and call the parameter `string $name`?

oe_content.install
```diff
@@ -0,0 +1,26 @@
...
+ */
+function oe_content_requirements($phase) {
+  $requirements = [];
+  if ($phase !== 'runtime') {
+    return $requirements;
+  }
+
+  if (!\Drupal::configFactory()->get('oe_content.settings')->get('provenance_uri')) {
+    $requirements['oe_content'] = [
+      'title' => t('OE Content Provenance URI'),
```
ademarco: We should always refer to the project name as `OpenEuropa` in UI strings.

tests/features/announcement.feature
```diff
@@ -1,9 +1,10 @@
-@api
+@api @run
```
ademarco: Remove `@run` leftover.

oe_content.install
```diff
@@ -0,0 +1,26 @@
...
+  if (!\Drupal::configFactory()->get('oe_content.settings')->get('provenance_uri')) {
+    $requirements['oe_content'] = [
+      'title' => t('OpenEuropa Content Provenance URI'),
+      'description' => t('The RDF entity provenance URI is not set.'),
+      'severity' => REQUIREMENT_ERROR,
+    ];
+  }
+
+  return $requirements;
+}
```
brummbar: Newline is required here.

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+  public function setDefaultProvenance(BeforeScenarioScope $scope): void {
+    $this->configContext->setConfig('oe_content.settings', 'provenance_uri', 'http://example.com');
+  }
+
+  /**
+   * Configures the Provenance URI.
+   *
+   * @Given the site Provenance URI is set to :uri
+   */
+  public function theSiteProvenanceUriIsSetTo($uri): void {
```
brummbar: The default suggested names from Behat are weird. Let's try to use a better terminology, like `setSiteProvenanceUri()`

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * User creation and login for creating RDF entities.
+   *
+   * A custom step that creates a user that has the permissions to create and
+   * view RDF entities of a give type.
+   *
+   * @Given I am logged in with a user that can create and view :bundle RDF entities
+   */
+  public function iAmLoggedInWithUserThatCanCreateAndViewRdfEntityTypes($bundle) {
```
brummbar: Let's follow the naming on the DrupalContext (even if it doesn't contain asserts...) 
```public function assertLoggedInWithRdfEntityTypePermissions($bundle): void {```

File: tests/Behat/RdfEntityContext.php [hunk 13]
```diff
@@ -0,0 +1,205 @@
...
+    // Login.
+    $this->login($user);
+  }
+
+  /**
+   * Creates a test RDF entity.
+   *
+   * @Given I create an :bundle RDF entity with the name :name
+   */
+  public function iCreateTestRdfEntityWithTheName($bundle, $name): void {
```
brummbar: `createRdfEntityWithLabel()`

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+
+    $this->createRdfEntity($values);
+  }
+
+  /**
+   * Tests that the created RDF entity has the correct provenance URI.
+   *
+   * @Then the Provenance URI of the RDF entity with the name :name should be :provenance_uri
+   */
+  public function theProvenanceUriOfTheRdfEntityWithTheNameShouldBe($name, $provenance_uri): void {
```
brummbar: `public function assertRdfEntityProvenanceUri(string $name, string $provenance_uri): void {`

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+      throw new \Exception('The RDF entity did not get created with the proper Provenance URI.');
+    }
+  }
+
+  /**
+   * Removes a RDF entity with a given name.
+   *
+   * @Then I delete the RDF entity with the name :name
+   */
+  public function iDeleteTheRdfEntityWithTheName($name): void {
```
brummbar: `public function deleteRdfEntityByLabel()`

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+    $this->createRdfEntity($values);
+  }
+
+  /**
+   * Tests that the created RDF entity has the correct provenance URI.
+   *
+   * @Then the Provenance URI of the RDF entity with the name :name should be :provenance_uri
+   */
+  public function theProvenanceUriOfTheRdfEntityWithTheNameShouldBe($name, $provenance_uri): void {
+    $entities = \Drupal::entityTypeManager()->getStorage('rdf_entity')->loadByProperties(['label' => $name]);
```
brummbar: We do this multiple times. Let's create `protected function getRdfEntityByLabel($label)`.

tests/Behat/RdfEntityContext.php
```diff
@@ -0,0 +1,205 @@
...
+namespace Drupal\Tests\oe_content\Behat;
+
+use Behat\Behat\Hook\Scope\AfterScenarioScope;
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
+use Drupal\DrupalExtension\Context\RawDrupalContext;
+
+/**
+ * Defines step definitions specifically for testing the RDF entities.
+ *
+ * This is needed due to issues with the Kernel testing of the RDF entity module
```
brummbar: Move to the test itself. The class is fine, it's just the test that is created to avoid the 8.6 issue.

tests/Behat/RdfEntityContext.php
upchuk: @brummbar I refactored this in the other ticket, will be done there.

PR #5, opened 2018-08-14
OPENEUROPA-1116: Import and create taxonomies needed for the demo

ademarco:


tests/features/vocabularies.feature
```diff
@@ -0,0 +1,16 @@
+@api @run
```
ademarco: Remove `@run`

oe_content.module
```diff
@@ -74,3 +75,31 @@ function oe_content_rdf_entity_access(EntityInterface $entity, $operation, Accou
   $provenance_uri = \Drupal::config('oe_content.settings')->get('provenance_uri');
   return $entity->get('provenance_uri')->value === $provenance_uri ? AccessResult::neutral() : AccessResult::forbidden();
 }
+
+/**
+ * Implements hook_entity_load().
+ */
+function oe_content_taxonomy_term_load(array $entities) {
+  // Taxonomy terms have a status field for handling publishing state. We need
+  // to default to TRUE for all terms.
```
ademarco: We have no choice now other than adding this to all loaded terms since the `rdf_taxonomy` module takes completely over (storage can be swapped on entity type, after all). Shall we maybe add a comment explaining that we will refactor this by switching from taxonomy to a custom entity type?

oe_content.module
```diff
@@ -74,3 +75,31 @@ function oe_content_rdf_entity_access(EntityInterface $entity, $operation, Accou
...
+  }
+}
+
+/**
+ * Implements hook_query_QUERY_TAG_alter() for the taxonomy_term_access tag.
+ */
+function oe_content_query_taxonomy_term_access_alter(AlterableInterface $query) {
+  if (!\Drupal::moduleHandler()->moduleExists('rdf_taxonomy')) {
+    return;
+  }
```
ademarco: This is not strictly necessary since this module depends from `rdf_taxonomy`. Maybe a reference to the todo note from the above comment would be enough?

runner.yml.dist
```diff
@@ -52,3 +52,10 @@ commands:
     - { task: "process", source: "phpunit.xml.dist", destination: "phpunit.xml" }
   setup:behat:
     - { task: "process", source: "behat.yml.dist", destination: "behat.yml" }
+  sparql:purge:
```
ademarco: We should add a comment above these saying something like the following:

```
# Use the following commands only if you can run Task Runner commands on your host machine.
```

README.md
```diff
@@ -132,3 +132,27 @@ To run the behat test:
...
+```
+$ ./vendor/bin/run sparql:import
+```
+
+Reset all data, i.e. run purge and import:
+
+
+```
+$ ./vendor/bin/run sparql:reset
+```
```
ademarco: The version below is more correct:

===============

Purge all data:

```
$ docker-compose exec sparql ./vendor/bin/robo purge
```

Or, if you can run commands on your host machine:

```
$ ./vendor/bin/run sparql:purge
```

Import default data:

```
$ docker-compose exec sparql ./vendor/bin/robo import
```

Or, if you can run commands on your host machine:

```
$ ./vendor/bin/run sparql:import
```

Reset all data, i.e. run purge and import:

```
$ docker-compose exec sparql ./vendor/bin/robo purge
$ docker-compose exec sparql ./vendor/bin/robo import
```

Or, if you can run commands on your host machine:

```
$ ./vendor/bin/run sparql:reset
```

ademarco:
Ok all good for me! @upchuk please approve it on my behalf as I cannot do it.

oe_content.module
claudiu-cristea: Use `hook_ENTITY_TYPE_storage_load()` instead of `hook_ENTITY_TYPE_load()`, because that also is caching the altered value. `hook_ENTITY_TYPE_load()` runs every time regardless if the items were retrieved from cache or not.

PR #11, opened 2018-08-27
OPENEUROPA 1117: Add term reference fields to the announcement bundle

upchuk:
### Description

Added some taxonomy fields + a missing location field.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_content_announcement/config/install/core.entity_form_display.rdf_entity.oe_announcement.default.yml
```diff
@@ -60,6 +66,41 @@ targetEntityType: rdf_entity
 bundle: oe_announcement
 mode: default
 content:
+  oe_announcement_department:
```
brummbar: We should order the fields in the same way done in the Excel file.

brummbar: Talked with team and for now the order is fine.

modules/oe_content_announcement/config/install/field.field.rdf_entity.oe_announcement.oe_announcement_location.yml
```diff
@@ -0,0 +1,18 @@
...
+    - field.storage.rdf_entity.oe_announcement_location
+    - rdf_entity.rdfentity.oe_announcement
+id: rdf_entity.oe_announcement.oe_announcement_location
+field_name: oe_announcement_location
+entity_type: rdf_entity
+bundle: oe_announcement
+label: Location
+description: ''
+required: true
+translatable: false
```
imanoleguskiza: This needs to be set to true

PR #18, opened 2018-09-04
OPENEUROPA-1199: Make sure RDF Entity module is compatible with Drupal core 8.6.x

ademarco:


ademarco:
This seems to fail since somehow the classmap autoload is not considered in the rdf_entity module, I find that adding it to the main composer.json of the component fixes the issue:

```
    "autoload": {
        "classmap": [
            "build/modules/contrib/rdf_entity/driver/sparql/Connection.php"
        ]
    },
```

PR #23, opened 2018-09-10
OPENEUROPA-1184: Making the event related links use a title and mapping to RDF.

upchuk:
### Description

Users should specify a title on the related link so we can output it cleanly. This PR makes it mandatory and also maps the value to RDF.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/features/event.feature
```diff
@@ -26,7 +27,7 @@ Feature: Event feature
     And I should see "My description"
     And I should see the link "legal expenses"
     And I should see "fr"
-    And I should see "http://example.com/my-event/more-info"
+    And I should see the link "My related link"
```
ademarco: I would still test that the link actually links to that URL, we could use a step like:
```php
<?php

class MinkContext extends RawMinkContext {

  /**
   * Check that a link pointing to a specific target.
   *
   * @Then I should see the link :link pointing to :url
   *
   * @throws \Exception
   *   If region or link within it cannot be found.
   */
  public function assertLinkRegion($link, $url) {
    $page = $this->getSession()->getPage();

    $result = $page->findLink($link);
    if (empty($result)) {
      throw new ExpectationException("No link '{$link}' on the page", $this->getSession());
    }

    $href = $result->getAttribute('href');
    if ($url != $href) {
      throw new ExpectationException("The link '{$link}' points to '{$href}'", $this->getSession());
    }

  }
```

PR #32, opened 2018-12-20
OPENEUROPA-1503: Add and configure content_lock module.

imanoleguskiza:
### Description

Prevent content being edited if it is already being edited by another user

### Change log

- Added: Content_lock module


upchuk:
@imanoleguskiza Please see my comment in the Jira ticket. This needs to go in another component.

PR #44, opened 2019-02-22
OPENEUROPA-1508 Create News Content Type

nagyad:


runner.yml.dist
```diff
@@ -23,8 +23,10 @@ drupal:
     - "./vendor/bin/drush config-set system.theme admin seven -y"
     - "./vendor/bin/drush config-set node.settings use_admin_theme 1 -y"
     # Enable the modules.
+    - "./vendor/bin/drush en oe_media -y"
```
upchuk: Since oe_content_news depends on this, we don't need to enable manually no?

nagyad: In this case, yes we don't need.

modules/oe_content_news/config/install/field.storage.node.oe_news_location.yml
```diff
@@ -0,0 +1,19 @@
...
+    - rdf_skos
+id: node.oe_news_location
+field_name: oe_news_location
+entity_type: node
+type: skos_concept_entity_reference
+settings:
+  target_type: skos_concept
+module: rdf_skos
+locked: false
+cardinality: -1
```
brummbar: The cardinality of this should be 1.

modules/oe_content_news/oe_content_news.info.yml
```diff
@@ -0,0 +1,31 @@
...
+core: 8.x
+dependencies:
+  - oe_content:oe_content
+  - oe_media:oe_media
+  - oe_media_avportal:oe_media_avportal
+
+config_devel:
+  - core.entity_form_display.node.oe_news.default
+  - core.entity_view_display.node.oe_news.default
+  - core.entity_view_display.node.oe_news.teaser
```
brummbar: I missed it in the page ticket, but let's remove the teaser display as we don't have the specifications for it nor any use case at the moment.

brummbar:
We remove the teaser from page in another PR.

PR #51, opened 2019-03-25
 OPENEUROPA-1667: Limit text field length

nagyad:


composer.json
```diff
@@ -8,24 +8,26 @@
     "require": {
         "drupal/core": "^8.6",
         "drupal/field_group": "^1.0",
+        "drupal/maxlength": "~1.0@beta",
         "easyrdf/easyrdf": "0.10.0-alpha.1 as 0.9.1",
         "openeuropa/rdf_skos": "dev-master",
-        "openeuropa/oe_media": "~0.2.0",
+        "openeuropa/oe_media": "~0.3",
```
ademarco: Why do we require "openeuropa/oe_media" here? Isn't the ticket about limiting text field length?

grumphp.yml.dist
```diff
@@ -1,2 +1,22 @@
 imports:
-  - { resource: vendor/openeuropa/code-review/dist/drupal-conventions.yml }
\ No newline at end of file
+  - { resource: vendor/openeuropa/code-review/dist/drupal-conventions.yml }
+parameters:
+  tasks.phpcs.ignore_patterns:
+    - build/
+    - node_modules/
```
ademarco: We don't have `node_modules` in this project, we can remove this line.

grumphp.yml.dist
```diff
@@ -1,2 +1,22 @@
...
+    - vendor/
+  tasks.phpcs.triggered_by:
+    - inc
+    - install
+    - module
+    - php
+    - theme
```
ademarco: We don't have .theme files here.

grumphp.yml.dist
```diff
@@ -1,2 +1,22 @@
...
+
+  extra_tasks:
+    phpparser:
+      ignore_patterns: %tasks.phpcs.ignore_patterns%
+      visitors:
+        declare_strict_types: ~
+      triggered_by: %tasks.phpcs.triggered_by%
+  extensions:
+    - OpenEuropa\CodeReview\ExtraTasksExtension
```
ademarco: Are we still using this in code-review 1.x? \cc @drupol

tests/features/news.feature
```diff
@@ -40,3 +40,19 @@ Feature: News content creation
     And I should see the link "financing"
     And I should see the link "European Patent Office"
     And I should see the AV Portal photo "Euro with miniature figurines" with source "//ec.europa.eu/avservices/avs/files/video6/repository/prod/photo/store/store2/4/P038924-352937.jpg"
+
+  @javascript
+  Scenario: Length limited fields are not allowing characters above the limit configured.
+    Given I am logged in as a user with the "create oe_news content, access content, edit own oe_page content, view published skos concept entities, create av_portal_photo media" permission
+    When I visit "the News creation page"
+    Then I should see the text "Content limited to 150 characters, remaining: 150"
+    And I should see the text "Content limited to 250 characters, remaining: 250"
```
ademarco: Can we contextualize these two steps? We can simply use Behat regions to target those specific fields, ideally we should have:

```gherkin
    Then I should see the text "Content limited to 150 characters, remaining: 150" in the "Title form element"
    And I should see the text "Content limited to 250 characters, remaining: 250" in the "Teaser form element"
```

tests/features/news.feature
```diff
@@ -40,3 +40,19 @@ Feature: News content creation
...
+    When I fill in "Title" with "My news"
+    And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
+    And I fill in "Teaser" with "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin eu hendrerit lacus, vitae bibendum odio. Fusce orci purus, hendrerit a magna at nullam. Text to remove"
+    And I fill in "Summary" with "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas felis leo, lobortis non eros in, consequat tempor est. Praesent sit amet sem eleifend, cursus arcu ac, eleifend nunc. Integer et orci sagittis, volutpat felis sit amet, tincidunt amet. Text to remove"
+    And I fill in "Body" with "Body text"
+    And I fill in "Subject" with "financing"
+    And I fill in "Author" with "European Patent Office"
+    And I press "Save"
+    Then I should not see "The text to remove."
```
ademarco: This is not clear, you are asserting that you could submit the form? Then you can assert on the content created message.

Anyway, according to the scenario description, we should test that content cannot be saved if max length is exceeded.

tests/features/page.feature
```diff
@@ -26,3 +26,16 @@ Feature: Page content creation
     And I should see "Teaser text"
     And I should see the link "financing"
     And I should see the link "European Patent Office"
+
+  @javascript
+  Scenario: Length limited fields are not allowing characters above the limit configured.
+    Given I am logged in as a user with the "create oe_page content, access content, edit own oe_page content, view published skos concept entities" permission
+    When I visit "the Page creation page"
+    Then I should see the text "Content limited to 150 characters, remaining: 150"
+    And I should see the text "Content limited to 250 characters, remaining: 250"
```
ademarco: Same as above.

tests/features/page.feature
```diff
@@ -26,3 +26,16 @@ Feature: Page content creation
...
+    When I fill in "Title" with "My page"
+    And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
+    And I fill in "Teaser" with "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin eu hendrerit lacus, vitae bibendum odio. Fusce orci purus, hendrerit a magna at nullam. Text to remove"
+    And I fill in "Summary" with "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas felis leo, lobortis non eros in, consequat tempor est. Praesent sit amet sem eleifend, cursus arcu ac, eleifend nunc. Integer et orci sagittis, volutpat felis sit amet, tincidunt amet. Text to remove"
+    And I press "Save"
+    Then I should not see "The text to remove."
```
ademarco: Same as above.

composer.json
nagyad: It was already there to support media fields.

tests/features/news.feature
nagyad: It will trim the extra characters. The only way to have more characters then the limit if we do copy paste. By typing you can't go over the limit, it will simply not type.

tests/features/page.feature
nagyad: It will trim the extra characters. The only way to have more characters then the limit if we do copy paste. By typing you can't go over the limit, it will simply not type.

grumphp.yml.dist
ademarco: Yes, that appears to be an outdated config file, it also does not check .module file, nor .php. let's should [use Drupal module template's one](https://github.com/openeuropa/drupal-module-template/blob/master/grumphp.yml.dist):

```yml
imports:
  - { resource: vendor/openeuropa/code-review/dist/drupal-conventions.yml }
parameters:
  tasks.phpcs.ignore_patterns:
    - build/
    - node_modules/
    - vendor/
  tasks.phpcs.triggered_by:
    - inc
    - install
    - module
    - php
    - theme
    - yml
```

grumphp.yml.dist
```diff
@@ -3,16 +3,11 @@ imports:
...
-    phpparser:
-      ignore_patterns: %tasks.phpcs.ignore_patterns%
-      visitors:
-        declare_strict_types: ~
-      triggered_by: %tasks.phpcs.triggered_by%
-  extensions:
-    - OpenEuropa\CodeReview\ExtraTasksExtension
+    - module
+    - php
+    - yml
```
voidtek: An empty line is needed.

grumphp.yml.dist
```diff
@@ -3,16 +3,11 @@ imports:
...
     - install
-
-  extra_tasks:
-    phpparser:
-      ignore_patterns: %tasks.phpcs.ignore_patterns%
-      visitors:
-        declare_strict_types: ~
-      triggered_by: %tasks.phpcs.triggered_by%
-  extensions:
-    - OpenEuropa\CodeReview\ExtraTasksExtension
```
voidtek: If we remove these tests, we should remove the dependencies on composer.json.
I propose to keep it.

grumphp.yml.dist
voidtek: The parser is only on `dev-master` of **code-review**. so, we should keep the **phpparser**, until the next version of the **code-review**.

PR #56, opened 2019-04-04
OPENEUROPA-1724: Make publication date a datetime field.

imanoleguskiza:
### Description

Make publication date field a datetime field.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_content_news/config/install/core.entity_form_display.node.oe_news.default.yml
```diff
@@ -30,7 +30,7 @@ content:
     third_party_settings: {  }
     region: content
   created:
-    type: datetime_timestamp
+    type: datetime_default
```
hernani: Added to wrong field.

PR #57, opened 2019-04-05
OPENEUROPA-1738: Add support for inline content linking

sergepavle:
### Description
We should provide editors with the possibility to link internal site content in a user-friendly. More over we need to make sure that links that are saved in any rich text fields on EWCMS is free of any Drupal-related tags/tokens/etc. this is important so that content we produced can be properly consumed by third-parties.

In order to do that we can leverage work done on OPENEUROPA-1737 and make sure that, when piking a content to link, a UUID-based canonical URL is saved in the rich text field, like in the example below:

{code}
<p>
Lorem ipsum <a href="http://my-site.ec.europa.eu/content/55c4b21a-7c9d-4736-b99b-f888ba057882">dolor sit amet</a>, consectetur adipiscing elit.
</p>
{code}

This will make sure that any third party consuming our content will have all links working, regardless of the current URL alias of the content.

Now, on our site we don't wont to print to the users the UUID code, but we want to actually show them the current URL alias, so we need to make sure that, at render time, we "find and replace" all those links with their related URLs.

Also, beware of the language negotiation, meaning we should catch also those:

{code}
<p>
Lorem ipsum <a href="http://my-site.ec.europa.eu/fr/content/55c4b21a-7c9d-4736-b99b-f888ba057882">dolor sit amet</a>, consectetur adipiscing elit.
</p>
{code}

Or those, if we will ever go with suffix again:

{code}
<p>
Lorem ipsum <a href="http://my-site.ec.europa.eu/content/55c4b21a-7c9d-4736-b99b-f888ba057882_fr">dolor sit amet</a>, consectetur adipiscing elit.
</p>
{code}

A possible solution is to use the Linkit module version 5.x (https://www.drupal.org/project/linkit) as it provides most of the things we need and a good level of customization.

We need to make sure that the site URL is known to the site somehow (this could be a service parameter or a simple configuration object, we will be able to override it via an ENV variables at build time in both cases).

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_content_persistent/src/ContentUuidResolver.php
```diff
@@ -89,4 +90,24 @@ public function getEntityByUuid(string $uuid, string $langcode = NULL): ?EntityI
...
+      }
+    }
+    return $supported_types;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isSupportedEntityType(EntityTypeInterface $entity_type): bool {
+    return in_array($entity_type->id(), $this->supportedStorages, TRUE);
```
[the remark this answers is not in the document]
sergepavle: Removed for now. This method was used for implemented for try to use entity link template ("purl"). Have found that exist some difficulties. I just forgot to remove in this class.

modules/oe_content_persistent/src/ContentUuidResolver.php
```diff
@@ -89,4 +90,24 @@ public function getEntityByUuid(string $uuid, string $langcode = NULL): ?EntityI
...
+        $supported_types[$entity_type_id] = $entity_type;
+      }
+    }
+    return $supported_types;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isSupportedEntityType(EntityTypeInterface $entity_type): bool {
```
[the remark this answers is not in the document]
sergepavle: Removed for now. This method was used for implemented for try to use entity link template ("purl"). Have found that exist some difficulties. I just forgot to remove in this class.

modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
```diff
@@ -0,0 +1,2 @@
+langcode: en
+inter_institutional_domain: 'https://data.ec.europa.eu/ewp/'
```
upchuk: Let's call this a base path, not domain. Since there is more than just the domain configured :)

modules/oe_content_persistent/src/Plugin/Filter/FilterPUrl.php
```diff
@@ -0,0 +1,89 @@
...
+   */
+  public function process($text, $langcode) {
+    $result = new FilterProcessResult($text);
+
+    $dom = Html::load($text);
+    $xpath = new \DOMXPath($dom);
+    foreach ($xpath->query('//a') as $node) {
+      $href = $node->getAttribute('href');
+
+      preg_match('/\/content\/([0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12})/i', $href, $matches);
```
upchuk: Can we put the regex strings as class constants so we can easier understand what we are preg_matching?

modules/oe_content_persistent/src/Form/PurlSettingsForm.php
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $config = $this->config('oe_content_persistent.settings');
+
+    $form['inter_institutional_domain'] = [
+      '#type' => 'url',
+      '#title' => $this->t('Inter institutional domain'),
+      '#default_value' => $config->get('inter_institutional_domain'),
```
upchuk:
```
'#description' => $this-t('The base URL to use for building persistent URLs'),
```

modules/oe_content_persistent/src/Plugin/Filter/FilterPUrl.php
```diff
@@ -0,0 +1,89 @@
...
+
+      // If there is a 'src' attribute, set it to the file entity's current
```
upchuk: `href` attribute no? I think this comment is copy-pasted from somewhere?

modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
sergepavle: done
didn't saw comment. currently I renamed to `inter_institutional_base_url`. Is this ok?

runner.yml.dist
```diff
@@ -23,11 +23,13 @@ drupal:
...
     - "./vendor/bin/drush config-set system.theme admin seven -y"
     - "./vendor/bin/drush config-set node.settings use_admin_theme 1 -y"
+    - "./vendor/bin/drush -y config-set system.performance css.preprocess 0"
+    - "./vendor/bin/drush -y config-set system.performance js.preprocess 0"
     # Enable the modules.
     - "./vendor/bin/drush en oe_content -y"
     - "./vendor/bin/drush en oe_content_page -y"
     - "./vendor/bin/drush en oe_content_news -y"
-    - "./vendor/bin/drush en oe_content_persistent -y"
+    - "./vendor/bin/drush en oe_content_persistent path -y"
```
upchuk: As discussed, this module is only needed for the test so you can enable it there.

tests/Behat/WysiwygContext.php
```diff
@@ -0,0 +1,354 @@
...
+      case $count === 0:
+        throw new \Exception("Could not find link '$node1_url' to relevant node.");
+
+      case $count > 1:
+        throw new \Exception("Multiple link '$node1_url' to relevant node found.");
+    }
+  }
+
+  /**
+   * Set alias of the node.
```
upchuk: There are methods such as this which do not belong in the WYSIWYG context right? Let's move them to the Feature or Drupal context.

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+    And I fill in "Title" with "News1"
+    And I enter "Teaser text" in the "Teaser" wysiwyg editor
+    And I enter "Summary text" in the "Summary" wysiwyg editor
+    And I enter "Body text" in the "Body" wysiwyg editor
+    And I fill in "Subject" with "financing"
+    And I fill in "Author" with "European Patent Office"
+    And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
+    And I press "Save"
+    Then I should see "News1"
+    #Create second node
```
upchuk:
```
# Create a second node
```

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+
+  @purl-linkit @javascript
+  Scenario: Make sure that persistent url work with linkit module.
+    Given I am logged in as a user with the "create oe_news content, access content, edit own oe_page content, view published skos concept entities, create av_portal_photo media, use text format base_html, view the administration theme" permission
+    And the following languages are available:
+      | languages |
+      | en        |
+      | fr        |
+    When I visit "the News creation page"
+    And I fill in "Title" with "News1"
```
upchuk:
```
News 1
```

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+    And I enter "Summary text" in the "Summary" wysiwyg editor
+    And I enter "Body text" in the "Body" wysiwyg editor
+    And I fill in "Subject" with "financing"
+    And I fill in "Author" with "European Patent Office"
+    And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
+    And I press "Save"
+    Then I should see "News1"
+    #Create second node
+    When I visit "the News creation page"
+    And I fill in "Title" with "News2"
```
upchuk: News 2

modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
sergepavle: renamed exactly to `base_url`

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+    And I enter "Teaser text" in the "Teaser" wysiwyg editor
+    And I enter "Summary text" in the "Summary" wysiwyg editor
+    And I insert link to "News1" through wysiwyg editor in "Body" field
```
upchuk:
```
And I insert a link to "News 1" in the "Body" field through the WYSIWYG editor
```

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+    And I fill in "Subject" with "financing"
+    And I fill in "Author" with "European Patent Office"
+    And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
+    And I press "Save"
+    Then I should see "News2"
+    # Check link to first node
+    And I should see link to "News1" node
+    # Check link to first node with alias
+    When I update alias of "News1" node to "/alias1"
+    Then I should see updated link to "News1" on "News2" page
```
upchuk: Then I should see the link "News 1" pointing to "/alias1"

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
     And I press "Save"
     # We assert that the extra characters are actually truncated from the end of the string.
     Then I should not see "The text to remove."
+
+  @purl-linkit @javascript
```
upchuk: Let's put it in a separate feature.

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+    And I enter "Teaser text" in the "Teaser" wysiwyg editor
+    And I enter "Summary text" in the "Summary" wysiwyg editor
+    And I insert link to "News1" through wysiwyg editor in "Body" field
+    And I fill in "Subject" with "financing"
+    And I fill in "Author" with "European Patent Office"
+    And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
+    And I press "Save"
+    Then I should see "News2"
+    # Check link to first node
+    And I should see link to "News1" node
```
upchuk: For this step, reuse the method `assertLinkWithHref()` once you got the URL of the node itself.

tests/features/news.feature
```diff
@@ -55,3 +55,41 @@ Feature: News content creation
...
+    # Check link to first node with alias
+    When I update alias of "News1" node to "/alias1"
+    Then I should see updated link to "News1" on "News2" page
+    # Check link to first node with alias for another language.
+    When I update alias of "News1" node to "/alias1_fr" for "French"
```
upchuk: You can make `updateNodeAlias` do this by passing an optional langcode param?

tests/Behat/FeatureContext.php
```diff
@@ -104,4 +108,147 @@ public function assertAvPortalPhoto(string $title, string $src): void {
...
+   * Check link to node.
+   *
+   * @param string $node_title
+   *   Title of the node.
+   * @param string|null $alias
+   *   Alias of the node.
+   *
+   * @throws \Drupal\Core\Entity\EntityMalformedException
+   *
+   * @Then I should see link to :node_title node
```
brummbar:
```suggestion
   * @Then I should see a link pointing to the :node_title node
```

tests/Behat/FeatureContext.php
```diff
@@ -104,4 +108,147 @@ public function assertAvPortalPhoto(string $title, string $src): void {
...
+   * @Then I should see the node :node_title pointing to :alias
```
brummbar:
```suggestion
   * @Then I should see a persistent link for the node :node_title pointing to :alias
```

tests/Behat/FeatureContext.php
```diff
@@ -104,4 +108,147 @@ public function assertAvPortalPhoto(string $title, string $src): void {
...
+   */
+  public function assertProcessedLink(string $node_title, $alias = NULL) {
+    $node1 = $this->getNodeByTitle($node_title);
```
brummbar: Replace $node1 with $node :)

PR #89, opened 2019-07-17
OPENEUROPA-2039: Update summary and body labels.

imanoleguskiza:
### Description

Update summary and body labels.

### Change log

- Added:
- Changed: Update summary and body labels.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_content_news/oe_content_news.post_update.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\field\Entity\FieldConfig;
+
+/**
+ * Update body and summary labels.
+ */
+function oe_content_news_post_update_field_labels(array &$sandbox): void {
```
nagyad: We have agreed that new upgrade path names has to include number from now on! Please fix all.

PR #90, opened 2019-07-23
 OPENEUROPA-2047: Add timeline field to policy content type.

nagyad:


modules/oe_content_policy/oe_content_policy.post_update.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types = 1);
+
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+
+/**
+ * Add missing permission for Validator.
+ */
+function oe_content_policy_post_update_00001_timeline_field(array &$sandbox): void {
```
sergepavle: I think according to last agreement about post update naming it should be `oe_content_policy_post_update_00001_timeline_field`

nagyad: They changed in the meantime.

composer.json
```diff
@@ -25,6 +25,7 @@
         "nikic/php-parser": "^3.1.5",
         "openeuropa/behat-transformation-context" : "~0.1",
         "openeuropa/code-review": "~1.0@beta",
+        "openeuropa/oe_multilingual": "dev-master",
```
upchuk: Since we already have a release, let's use that one.

modules/oe_content_timeline_field/src/Plugin/Field/FieldWidget/TimelineFieldWidget.php
```diff
@@ -27,9 +26,6 @@ class TimelineFieldWidget extends WidgetBase implements WidgetInterface {
    * {@inheritdoc}
    */
   public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
-    $element += [
```
upchuk: Why arent you validating anymore?

tests/Behat/FeatureContext.php
```diff
@@ -250,4 +251,48 @@ protected function getLangcodeByName(string $language_name): string {
     throw new \Exception("Language name '$language_name' is not valid.");
   }
 
+  /**
+   * Step to fill in multi value fields with columns.
+   *
+   * @Given I fill in :column with :value in the :row :field field element
+   */
+  public function iFillInWithInTheFieldElement($column, $value, $row, $field) {
```
upchuk: `fillInMultivalueField`

tests/Behat/FeatureContext.php
```diff
@@ -250,4 +251,48 @@ protected function getLangcodeByName(string $language_name): string {
...
+  }
+
+  /**
+   * Finds the table that holds a multiple columned field.
+   *
+   * @param string $field
+   *   The field.
+   *
+   * @return \Behat\Mink\Element\NodeElement
+   *   The selector.
```
upchuk: The table element.

composer.json
nagyad: It's require-dev, why?

modules/oe_content_timeline_field/src/Plugin/Field/FieldWidget/TimelineFieldWidget.php
nagyad: That was for a specific field element. Since we don't need that anymore because if any of the fields are filled in is valid.

tests/features/policy.feature
```diff
@@ -18,15 +23,38 @@ Feature: Policy content creation
     And I fill in "Legacy link" with "http://example.com"
     And I fill in "Navigation title" with "Navi title"
     And I fill in "Short title" with "Shorter title"
+    And I fill in "Label" with "Label 1" in the "first" "Timeline" field element
```
dxvargas: Shouldn't label be removed from widget?

nagyad: We hide it in oe_theme in order to facilitate the ECLv2 implementation.

PR #97, opened 2019-08-26
OPENEUROPA-2188: Upgrade field_group to 3.0-rc1.

nagyad:


phpunit.xml.dist
```diff
@@ -7,6 +7,7 @@
     <env name="SIMPLETEST_BASE_URL" value="${drupal.base_url}"/>
     <env name="SIMPLETEST_DB" value="mysql://${drupal.database.user}:${drupal.database.password}@${drupal.database.host}:${drupal.database.port}/${drupal.database.name}"/>
     <env name="SIMPLETEST_SPARQL_DB" value="sparql://${drupal.sparql.host}:${drupal.sparql.port}/"/>
+    <env name="MINK_DRIVER_ARGS_WEBDRIVER" value='["${selenium.browser}", null, "${selenium.host}:${selenium.port}/wd/hub"]'/>
```
voidtek: this part is out of the ticket scope, this code can be added during the implementation of kernel test, in the future.
Now it's not needed.

composer.json
```diff
@@ -7,7 +7,7 @@
     "prefer-stable": true,
     "require": {
         "drupal/core": "^8.7",
-        "drupal/field_group": "^1.0",
+        "drupal/field_group": "^3.0.0-rc1",
```
brummbar: We actually need to drop this dependency as it's a leftover. We will move it to the ewcms.

PR #103, opened 2019-09-06
OPENEUROPA-2262: Remove name mappings on image and document media types.

imanoleguskiza:
### Description

Given an existing media image, if a user edits it and changes one of the values of the source field (alternative text in this case) the image will get its title replaced with the file name.

We need to go through all the media types we created and remove any field mapping that is actually not provided from the source OR that we don't want provided from the source

### Change log

- Added:
- Changed: Remove name mappings on image and document media types.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_content.post_update.php
+++ b/oe_content.post_update.php
@@ -1,0 +1,27 @@
+<?php
+
+/**
+ * @file
+ * Post update functions for OpenEuropa Content module.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\media\Entity\MediaType;
+
+/**
+ * Remove the mapping of name to title on documents and image media types.
+ */
+function oe_content_post_update_00001(): void {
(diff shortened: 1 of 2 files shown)


oe_content.post_update.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\media\Entity\MediaType;
+
+/**
+ * Remove the mapping of name to title on documents and image media types.
+ */
+function oe_content_post_update_0001(): void {
```
sergepavle: as I remember we agreed to have 5 digits.
```suggestion
function oe_content_post_update_00001(): void {
```

nagyad:
Update hook moved to EWCMS.

PR #115, opened 2019-10-07
OPENEUROPA-2224: Add support of Country corporate vocabulary.

sergepavle:
### Description

Create the country corporate vocabulary in rdf skos. 

Ensure it gets added to the remote triple store self-update thing.

See attachment for the values.
### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_content.post_update.php
+++ b/oe_content.post_update.php
@@ -54,5 +54,20 @@ function oe_content_post_update_00001(): void {
  * Add Country corporate vocabulary.
  */
 function oe_content_post_update_00002(): void {
-  \Drupal::service('oe_content.op_skos_setup')->setup();
+  $config = \Drupal::configFactory()->getEditable('rdf_skos.graphs');
+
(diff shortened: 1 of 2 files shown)


oe_content.post_update.php
```diff
@@ -49,3 +49,10 @@ function oe_content_post_update_00001(): void {
     }
   }
 }
+
+/**
+ * Add Country corporate vocabulary.
+ */
+function oe_content_post_update_00002(): void {
+  \Drupal::service('oe_content.op_skos_setup')->setup();
```
brummbar: We cannot assume that the module users didn't add another graph already. Let's change this to only add the country vocabulary.

sergepavle:
> This should be against EPIC-Event branch. Is there a specific reason you want this in master? @sergepavle

Yes agree. Fixed.

PR #116, opened 2019-10-08
OPENEUROPA-2216: Add gallery field to event.

nagyad:


tests/features/event.feature
```diff
@@ -153,10 +156,13 @@ Feature: Event content creation
     # Organiser field group.
     When I press "Organiser"
     Then I fill in "Organiser name" with "Organiser name"
+    # Media field group.
+    When I press "Media"
+    And I fill in "Use existing media" with "Visit by Federica Mogherini, Vice-President of the EC, and Johannes Hahn, Member of the EC, to Romania" in the "Gallery" region
```
sergepavle: Probably better to use `Then` as in other places.

PR #121, opened 2019-10-11
OPENEUROPA-2212: Add event profile reference to event.

nagyad:


modules/oe_content_event/config/install/core.entity_form_display.event_profile.oe_event_partner.default.yml
```diff
@@ -57,4 +44,6 @@ content:
     third_party_settings: {  }
     type: link_default
     region: content
-hidden: {  }
+hidden:
```
imanoleguskiza: The same changes should probably apply to other entities that are used in inline entity forms? Not a requirement but maybe worth a shot.

tests/features/event.feature
```diff
@@ -218,6 +228,11 @@ Feature: Event content creation
     And I should see the text "Rue belliard 28"
     And I should see the text "1000 Brussels"
     And I should see the text "Belgium"
+    # Event partner values.
+    And I should see the text "Name of the event partner"
+    And I should see an "AV portal photo of the Event partner" element
```
imanoleguskiza: Should we not check that the image is there, not just the wrapper div?

nagyad: Not yet, the theming ticket is still to be done. There we will take care of checking these. This is just a smoke test, that the field is actually output in any way.

modules/oe_content_event/config/install/core.entity_form_display.event_profile.oe_event_partner.default.yml
nagyad: It might be, for now I just applied here within the scope of the ticket. As soon as we receive the requirements for the back-end form, we will change the Venue as well.

PR #125, opened 2019-10-18
OPENEUROPA-2395: Add date translations.

nagyad:


nagyad:
We discussed with @kugta that not every terms are translated and we ship what we can. By the way for now we are not even using these, I just added them just in case.

PR #126, opened 2019-10-21
OPENEUROPA-2211: Add languages reference to event.

nagyad:


modules/oe_content_event/config/install/field.storage.event_profile.phone_number.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_content_event
+id: event_profile.phone_number
+field_name: phone_number
+entity_type: event_profile
+type: string
```
sergepavle: @nagyad 
Do we have any reason why we don't use telephone field type from Drupal core for this field?

nagyad: Yes, it was defined as text field by the analysts. I asked, but they don't want to bother with the constraints of different national numbers and length etc... This is an editorial input and not user input so it's up to the editor to ensure that the number is correct.

nagyad: image

PR #136, opened 2019-11-15
Extract and reuse anchor.

haringsrob:
### Description

Currently anchors are removed from the new url. This change reuses it.

### DIFF
--- a/modules/oe_content_persistent/src/Plugin/Filter/FilterPurl.php
+++ b/modules/oe_content_persistent/src/Plugin/Filter/FilterPurl.php
@@ -109,11 +109,21 @@ class FilterPurl extends FilterBase implements ContainerFactoryPluginInterface {
           continue;
         }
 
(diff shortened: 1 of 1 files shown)


modules/oe_content_persistent/src/Plugin/Filter/FilterPurl.php
```diff
@@ -109,11 +109,21 @@ public function process($text, $langcode) {
           continue;
         }
 
+        // If an anchor is set on the source url, we extract it here.
+        $anchor = '';
+        if (strpos($href, '#') !== FALSE) {
+          $explodedHref = explode('#', $href);
+          if (isset($explodedHref[1]) && !empty($explodedHref[1])) {
```
haringsrob: I think isset can be replaced by empty, but I know I had some issues with this in the past. So for safety I included both.

brummbar:
Fixed by #143 .

PR #138, opened 2019-11-22
OPENEUROPA-2514: Make sure that skos concepts are there for event.

nagyad:


nagyad:
We had to fix the version of field_group because the rc2 version causes a bug with tabs when loading the form. Specifically, the last or second tab is focused instead of the first tab. This was detected on EWCMS where tabs are actively used.

PR #139, opened 2019-11-25
OPENEUROPA-2532: Set default value for event status.

nagyad:
Branched out of https://github.com/openeuropa/oe_content/pull/138

### DIFF
--- a/modules/oe_content_event/oe_content_event.module
+++ b/modules/oe_content_event/oe_content_event.module
@@ -31,16 +31,22 @@ function oe_content_event_form_node_form_alter(&$form, FormStateInterface $form_
   $form['is_organiser_internal'] = [
     '#type' => 'checkbox',
     '#title' => t('Organiser is internal'),
-    '#default_value' => $form_state->getValue('oe_event_organiser_internal') ? 1 : 0,
+    '#default_value' => !empty($form['oe_event_organiser_name']['widget'][0]['value']['#default_value']) ? 0 : 1,
   ];
   $form['oe_event_organiser_internal']['#states'] = [
     'visible' => [
-      ':input[name="is_organiser_internal"]' => ['checked' => TRUE],
+      [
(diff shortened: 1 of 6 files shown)


tests/features/event.feature
```diff
@@ -62,10 +62,10 @@ Feature: Event content creation
     # The organiser group is opened by default.
     And I should see the text "Organiser"
     And I should see the text "Organiser is internal"
-    And I should see the text "Organiser name"
-    When I check "Organiser is internal"
-    Then I should see the text "Internal organiser"
-    And I should not see the text "Organiser name"
+    And the "Internal organiser field" is visible
```
imanoleguskiza: We should add an extra step "And I should not see the text "Organiser name"

nagyad:
@imanoleguskiza I see your point and it makes sense, however they don't do it. I mean form states. And they reason might be because you click but it might not be intentional and you don't want to loose the value that was there from before, need reload to if you don't remember what was it exactly.

imanoleguskiza:
I think we could check whether the checkbox is checked or not and then both validate and save the values depending on that, but since it is not a hard requirement, I will let the next reviewer give us their opinion.

modules/oe_content_event/oe_content_event.module
```diff
@@ -31,16 +31,22 @@ function oe_content_event_form_node_form_alter(&$form, FormStateInterface $form_
   $form['is_organiser_internal'] = [
     '#type' => 'checkbox',
     '#title' => t('Organiser is internal'),
-    '#default_value' => $form_state->getValue('oe_event_organiser_internal') ? 1 : 0,
+    '#default_value' => !empty($form['oe_event_organiser_name']['widget'][0]['value']['#default_value']) ? 0 : 1,
```
dxvargas: Better:
```empty($form['oe_event_organiser_name']['widget'][0]['value']['#default_value']) ? 1 : 0```

PR #140, opened 2019-11-28
OPENEUROPA-2221: Add Venue default formatter settings.

nagyad:


ademarco:
Closing as we don't have the EntityVenue entity anymore. See #149

PR #149, opened 2020-01-03
OPENEUROPA-2668: Configure event reference fields for new content entities.

nagyad:


tests/features/event.feature
```diff
@@ -6,7 +6,7 @@ Feature: Event content creation
 
   @javascript
   Scenario: Fields on the event content creation forms should be grouped logically.
-    Given I am logged in as a user with the "create oe_event content, access content, edit own oe_event content, view published skos concept entities, administer event venue entities, administer event profile entities" permission
+    Given I am logged in as a user with the "create oe_event content, access content, edit own oe_event content, view published skos concept entities, manage custom content entities" permission
```
ademarco: Question here:
```
Given I am logged in as a user with the "create oe_event content, access content, edit own oe_event content, view published skos concept entities, manage custom content entities" permission
```

Do users need to have the `manage custom content entities` permission in order to create corporate entities via an inline form on the event content type? If so we might need to break down that catch all permission into separate ones

modules/oe_content_event/oe_content_event.module
```diff
@@ -7,76 +7,44 @@
...
+function oe_content_event_node_presave(EntityInterface $entity) {
+  // Make sure that we don't save both concurrent field values.
+  if ($entity->bundle() === 'oe_event') {
+    $value = $entity->get('oe_event_organiser_is_internal')->value;
+    if ($value) {
+      $entity->get('oe_event_organiser_name')->setValue(NULL);
+    }
+    elseif ($value === 0) {
+      $entity->get('oe_event_organiser_internal')->setValue(NULL);
+    }
```
ademarco: This looks strange, since we know that the value is either 1 or 0 can't we simply do the following?
```php
    if ($entity->get('oe_event_organiser_is_internal')->value === 1) {
      $entity->get('oe_event_organiser_name')->setValue(NULL);
    }
    else {
      $entity->get('oe_event_organiser_internal')->setValue(NULL);
    }
```

modules/oe_content_event/src/EventUtilities.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_content_event;
+
+/**
+ * Provides generic event helper methods.
+ */
+class EventUtilities {
```
ademarco: Since this is generic, and not bound to the event content type only, please make it a static class, rename it as below  and move this to the following location:

```php
Drupal\oe_content_event\Utilities

/**
 * Provides helper methods to manipulate content forms.
 */
class ContentFormUtilities {
```

modules/oe_content_event/src/EventUtilities.php
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Helper method to apply the toggle states on two form fields.
+   *
+   * @param array $form
+   *   The drupal form array.
+   * @param string $toggle_field
+   *   The toggle field name.
```
ademarco: Since we need this to be a checkbox let's make it clear here:
```
   * @param string $checkbox_field
   *   The toggle field name.
```

modules/oe_content_event/src/EventUtilities.php
```diff
@@ -0,0 +1,51 @@
...
+   * @param string $field1
+   *   The dependent field name.
+   * @param string $field2
+   *   The dependent field name.
+   */
+  public static function applyToggleStatesToFormFields(array &$form, string $toggle_field, string $field1, string $field2): void {
```
ademarco: Please rename into:
```php
public static function toggleFieldsWithCheckbox(...)
```

modules/oe_content_event/oe_content_event.module
nagyad: This only works if the oe_event_organiser_is_internal is a required field.

tests/Behat/FeatureContext.php
```diff
@@ -349,4 +298,81 @@ public function assertVisibility($element): void {
...
+
+      $media = \Drupal::service('entity_type.manager')
+        ->getStorage('media')->create([
+          'bundle' => 'av_portal_photo',
+          'oe_media_avportal_photo' => $photo_id,
+          'uid' => 0,
+          'status' => 1,
+        ]);
+
+      $media->save();
```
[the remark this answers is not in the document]
nagyad: After researching I came up with a similar solution like in RawDrupalContext.

tests/Behat/MediaCreationContext.php
```diff
@@ -0,0 +1,139 @@
...
+   * | name 2 | file 1 |
+   * |   ...  |   ...  |
+   * // phpcs:enable
+   */
+  public function createMediaDocuments(TableNode $file_table): void {
+    // Retrieve the url table from the test scenario.
+    $files = $file_table->getRows();
+
+    foreach ($files as $file_properties) {
+      $file = file_save_data(file_get_contents(drupal_get_path('module', 'oe_content') . '/tests/fixtures/' . $file_properties[1]), 'public://' . $file_properties[1]);
```
ademarco: We have to use here the `files_path` property in the module's `behat.yml.dist`, and not the module's path, as this is supposed to be a reusable behat context.

You can access it by calling `$this->getMinkParameter('files_path')`:

```php
$file = file_save_data(file_get_contents($this->getMinkParameter('files_path') . $file_properties[1]), 'public://' . $file_properties[1]);
```

PR #150, opened 2020-01-10
OPENEUROPA-2640: Swap date fields to dateitem_range.

nagyad:
https://github.com/openeuropa/oe_content/pull/149 to be merged at first!


tests/Behat/FeatureContext.php
```diff
@@ -72,6 +104,29 @@ public function findDateFields($field) {
...
+    $field_selectors = $this->getSession()->getPage()->findAll('css', '.field--widget-daterange-default');
+    $field_selectors = array_filter($field_selectors, function ($field_selector) use ($field) {
+      return $field_selector->has('named', ['content', $field]);
+    });
+    if (empty($field_selectors)) {
+      throw new \Exception("Date range field {$field} was not found.");
+    }
+    return $field_selectors;
+  }
+
```
ademarco: We should move both of these steps in reusable context, so that we can re-use them in, for example, EWCMS Behat tests. We can call the context `DateFieldContext`.

PR #151, opened 2020-01-17
OPENEUROPA-2730: Update common fields label and description.

22Alexandra:
### Description

Update common fields label and description.

### Change log

- Added: fields description
- Changed: title, oe_content_short_title, oe_content_legacy_link labels


tests/features/page.feature
```diff
@@ -34,10 +34,25 @@ Feature: Page content creation
...
-    When I fill in "Title" with "My page"
+    When I fill in "Page title" with "My page"
     And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
     And I fill in "Teaser" with "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin eu hendrerit lacus, vitae bibendum odio. Fusce orci purus, hendrerit a magna at nullam. Text to remove"
     And I fill in "Introduction" with "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas felis leo, lobortis non eros in, consequat tempor est. Praesent sit amet sem eleifend, cursus arcu ac, eleifend nunc. Integer et orci sagittis, volutpat felis sit amet, tincidunt amet. Text to remove"
     And I press "Save"
     # We assert that the extra characters are actually truncated from the end of the string.
     Then I should not see "The text to remove."
+
+  Scenario: Fields have the correct description.
```
upchuk: This scenario is highly duplicared. Can we make a separate test feature for these kinds of things where we run through each content type? They are the common fields after all.

oe_content.module
```diff
@@ -23,11 +23,12 @@ function oe_content_entity_base_field_info(EntityTypeInterface $entity_type) {
...
 
   $fields['oe_content_short_title'] = BaseFieldDefinition::create('string')
-    ->setLabel(t('Short title'))
+    ->setLabel(t('Alternative title'))
+    ->setDescription(t('Use this field to create an alternative title for use in the URL and in list views. If the page title is longer than 60 characters, you can add a shorter title here.'))
     ->setRequired(FALSE)
     ->setTranslatable(TRUE)
     ->setRevisionable(TRUE)
-    ->setSetting('max_length', 255)
+    ->setSetting('max_length', 170)
```
upchuk: Shouldn't we handle this like we handle the title max length? With the maxlength module? Like we limit [for the title field](https://github.com/openeuropa/oe_content/blob/master/modules/oe_content_page/config/install/core.entity_form_display.node.oe_page.default.yml#L143). What you did here only controls the max size that fits in the DB. And in any case, we should not change that from 255 because it will become impossible to change back if we ever need to. We don't change any other right?

nagyad:
Looks good, but in the ticket I see max_length 170 chars setting to be applied for Alternative title. Why is that missing? @22Alexandra

PR #164, opened 2020-02-24
OPENEUROPA-2221: Add entity type default theme functions and event wrapper class

ademarco:


src/EntityWrapperBase.php
```diff
@@ -0,0 +1,50 @@
...
+   */
+  abstract protected function getWrappedEntityId(): string;
+
+  /**
+   * Get wrapped entity bundle machine name.
+   *
+   * @return string
+   *   Wrapped entity bundle machine name.
+   */
+  abstract protected function getWrappedEntityBundle(): string;
```
upchuk: What if the entity type does not have bundles? The `bundle` entity key may be missing? Such as User

src/EntityWrapperBase.php
```diff
@@ -0,0 +1,50 @@
...
+    $this->entity = $entity;
+  }
+
+  /**
+   * Get wrapped entity ID machine name.
+   *
+   * @return string
+   *   Wrapped entity ID machine name.
+   */
+  abstract protected function getWrappedEntityId(): string;
```
upchuk: These abstract methods look super nice to be public and on a generic interface. The wrapper can be inspected as to this information from the outside as well to see what kind of entity they wrap without having to dig out the entity. Which as it seems, you cannot anymore once you wrapped it. So that should also be possible.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
```diff
@@ -114,4 +106,36 @@ protected function validateOrganiserGroupFields(Constraint $constraint, NodeInte
     }
   }
 
+  /**
+   * Helper function to provide violation on a set of "Registration" fields.
```
upchuk: `Validates the event registration fields`

modules/oe_content_event/tests/src/Kernel/EventBusinessLogicTest.php
```diff
@@ -5,67 +5,22 @@
...
-    oe_content_install();
-  }
-
-  /**
-   * Test the Organisation fields.
+   * An organiser can either be a custom string or a reference to a corporate
+   * vocabulary, depending from the value of `oe_event_organiser_is_internal`.
+   *
+   * This tests that, if one is set, the other is always not, depending
+   * whereas the organiser is marked as internal or not.
```
upchuk: whether not whereas

src/Utility/ContentFormHandler.php
```diff
@@ -2,24 +2,24 @@
...
 declare(strict_types = 1);
 
-namespace Drupal\oe_content_event\Utilities;
+namespace Drupal\oe_content\Utility;
 
 /**
  * Provides helper methods to manipulate content forms.
  */
-class ContentFormUtilities {
+class ContentFormHandler {
```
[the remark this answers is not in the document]
ademarco: `ContentFormUtilities` it is

modules/oe_content_event/src/EventNodeWrapperInterface.php
```diff
@@ -0,0 +1,167 @@
...
+   */
+  public function isPostponed(): bool;
+
+  /**
+   * Check whereas the event has registration.
+   *
+   * @return bool
+   *   Whereas the event has registration.
+   */
+  public function hasRegistration(): bool;
```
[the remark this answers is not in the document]
ademarco: After discussing it with @upchuk we will keep it as it is.

modules/oe_content_event/tests/src/Kernel/EventBusinessLogicTest.php
ademarco: I've moved this block to the actual class, as it belongs there rather than in the theme.

modules/oe_content_event/src/EventNodeWrapperInterface.php
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Check whether the event registration is open.
+   *
+   * We consider the registration open if the event:
+   *
+   * 1. has a registration URL set;
+   * 2. and the event is not postponed;
+   * 3. and the event is not cancelled;
+   * 4. is withing the active registration period, only if dates are set.
```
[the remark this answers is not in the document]
nagyad: fixed in the commit above

ademarco:
@22Alexandra approved on @upchuk 's behalf.

PR #176, opened 2020-03-16
OPENEUROPA-2604: Drupal 8 recommended project.

upchuk:


composer.json
```diff
@@ -27,9 +28,12 @@
         "openeuropa/drupal-core-require-dev": "^8.7",
         "openeuropa/task-runner": "~1.0@beta",
         "phpunit/phpunit": "~6.0",
-        "symfony/dom-crawler": "~3.0",
+        "symfony/dom-crawler": "~3.4",
         "drupaltest/behat-traits": "dev-8.x-1.x"
     },
+    "_readme": [
+        "We use a minimum version of Pathauto 1.6 until a new release of oe_multilingual includes it. Then it can be removed."
+    ],
```
ademarco: Is this still needed? I mean, for lower deps we get pathauto 1.5 and drupal 8.7 and higher deps pathauto 1.6 and drupal 8.8. right?

composer.json
```diff
@@ -6,7 +6,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "drupal/core": "^8.7",
+        "drupal/core-recommended": "^8.7",
```
ademarco: I wouldn't add this here, we can wait for 8.8 to do that?

upchuk: It goes back to 8.7 releases as well.

PR #178, opened 2020-03-24
OPENEUROPA-3008: Map status field for tests in event.

nagyad:


tests/Behat/Content/Venue/DefaultVenueContext.php
```diff
@@ -27,10 +27,17 @@ public function alterVenueFields(BeforeParseEntityFieldsScope $scope): void {
...
     ];
 
     foreach ($scope->getFields() as $key => $value) {
       if (isset($mapping[$key])) {
+        if ($key === 'Published') {
+          $scope->addFields([
+            $mapping[$key] => (int) ($value === 'Yes'),
+          ])->removeField($key);
+          continue;
+        }
```
ademarco: Let's use a switch here as we did in `EventContentContext` so we keep consistency.

tests/Behat/Content/Contact/PressContactContext.php
```diff
@@ -28,10 +28,17 @@ public function alterPressContactFields(BeforeParseEntityFieldsScope $scope): vo
...
 
     foreach ($scope->getFields() as $key => $value) {
       if (isset($mapping[$key])) {
+        if ($key === 'Published') {
+          $scope->addFields([
+            $mapping[$key] => (int) ($value === 'Yes'),
+          ])->removeField($key);
+          continue;
+        }
         $scope->renameField($key, $mapping[$key]);
```
ademarco: Let's use a switch here as we did in `EventContentContext` so we keep consistency.

PR #187, opened 2020-04-03
OPENEUROPA-3050: Add link list source filter for events

imanoleguskiza:
### Description

Dynamic link lists of event nodes should have an additional filter that allows to chose if the event displayed should be past or upcoming ones.

### Change log

- Added: New link list source filter to filter events by date.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_content_event/oe_content_event.info.yml
+++ b/modules/oe_content_event/oe_content_event.info.yml
@@ -19,7 +19,6 @@ dependencies:
   - field_group:field_group
   - inline_entity_form:inline_entity_form
   - entity_reference_revisions:entity_reference_revisions
-  - oe_time_caching:oe_time_caching
 
 config_devel:
   install:
--- a/modules/oe_content_event/oe_content_event.info.yml
+++ b/modules/oe_content_event/oe_content_event.info.yml
@@ -19,6 +19,7 @@ dependencies:
   - field_group:field_group
   - inline_entity_form:inline_entity_form
   - entity_reference_revisions:entity_reference_revisions
+  - oe_time_caching:oe_time_caching
 
 config_devel:
   install:
--- a/modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
+++ b/modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
@@ -87,8 +87,8 @@ class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implement
       '#empty_value' => 'all',
       '#empty_option' => $this->t('Show all'),
       '#options' => [
-        'past' => $this->t('Past events'),
-        'upcoming' => $this->t('Upcoming events'),
+        self::PAST => $this->t('Past events'),
+        self::UPCOMING => $this->t('Upcoming events'),
       ],
     ];
     return $form;
--- a/modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
+++ b/modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
@@ -1,0 +1,98 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_content_event\Plugin\InternalLinkSourceFilter;
+
+use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
+use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\Core\Entity\Query\QueryInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
+use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterInterface;
+use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterPluginBase;
+
+/**
+ * Event link source filter class.
+ *
+ * @InternalLinkSourceFilter(
+ *   id = "event_link_source_filter",
+ *   label = @Translation("Event Link Source Filter"),
+ *   description = @Translation("Filters for link lists using events as the internal source."),
+ *   entity_types = {
+ *     "node" = {
+ *       "oe_event",
+ *     },
+ *   },
+ * )
+ */
+class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implements InternalLinkSourceFilterInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
+      'time' => 'upcoming',
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isApplicable(string $entity_type, string $bundle): bool {
+    $applicable_entity_types = $this->pluginDefinition['entity_types'];
+    if (isset($applicable_entity_types[$entity_type])) {
+      $applicable_bundles = $applicable_entity_types[$entity_type];
+      if (in_array($bundle, $applicable_bundles)) {
+        return TRUE;
+      }
+    }
+    return FALSE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function apply(QueryInterface $query, array $context, RefinableCacheableDependencyInterface $cacheability): void {
+    $now = new DrupalDateTime('now');
+    $now->setTimezone(new \DateTimeZone(DateTimeItemInterface::STORAGE_TIMEZONE));
+    switch ($this->getConfiguration()['time']) {
+      case 'past':
+        $query->condition('oe_event_dates.end_value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), "<=");
+        $query->sort('oe_event_dates.end_value', 'DESC');
+        break;
+
+      case 'upcoming':
+        $query->condition('oe_event_dates.value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), ">=");
+        $query->sort('oe_event_dates.value', 'ASC');
+        break;
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form['time'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Choose whether to show past or upcoming events.'),
+      '#default_value' => $this->getConfiguration()['time'] ?? 'upcoming',
+      '#empty_value' => 'all',
+      '#empty_option' => $this->t('Show all'),
+      '#options' => [
+        'past' => $this->t('Past events'),
+        'upcoming' => $this->t('Upcoming events'),
+      ],
+    ];
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $this->configuration['time'] = $form_state->getValue('time');
+  }
+
+}
--- a/modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
+++ b/modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
@@ -29,11 +29,21 @@ use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterPluginBase;
 class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implements InternalLinkSourceFilterInterface {
 
   /**
+   * Option for upcoming events.
+   */
+  const UPCOMING = 1;
+
+  /**
+   * Option for past events.
+   */
+  const PAST = -1;
+
+  /**
    * {@inheritdoc}
    */
   public function defaultConfiguration() {
     return [
-      'time' => 'upcoming',
+      'time' => self::UPCOMING,
     ];
   }
 
@@ -41,12 +51,8 @@ class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implement
    * {@inheritdoc}
    */
   public function isApplicable(string $entity_type, string $bundle): bool {
-    $applicable_entity_types = $this->pluginDefinition['entity_types'];
-    if (isset($applicable_entity_types[$entity_type])) {
-      $applicable_bundles = $applicable_entity_types[$entity_type];
-      if (in_array($bundle, $applicable_bundles)) {
-        return TRUE;
-      }
+    if (isset($this->pluginDefinition['entity_types'][$entity_type]) && in_array($bundle, $this->pluginDefinition['entity_types'][$entity_type])) {
+      return TRUE;
     }
     return FALSE;
   }
@@ -58,13 +64,13 @@ class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implement
     $now = new DrupalDateTime('now');
     $now->setTimezone(new \DateTimeZone(DateTimeItemInterface::STORAGE_TIMEZONE));
     switch ($this->getConfiguration()['time']) {
-      case 'past':
-        $query->condition('oe_event_dates.end_value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), "<=");
+      case self::PAST:
+        $query->condition('oe_event_dates.end_value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), "<");
         $query->sort('oe_event_dates.end_value', 'DESC');
         break;
 
-      case 'upcoming':
-        $query->condition('oe_event_dates.value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), ">=");
+      case self::UPCOMING:
+        $query->condition('oe_event_dates.value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), ">");
         $query->sort('oe_event_dates.value', 'ASC');
         break;
     }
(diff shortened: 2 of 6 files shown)


modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,98 @@
...
+ * )
+ */
+class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implements InternalLinkSourceFilterInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
+      'time' => 'upcoming',
```
sergepavle, line 36: As an option, it could be class constants UPCOMING and PAST with values 1 and -1 accordingly.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function apply(QueryInterface $query, array $context, RefinableCacheableDependencyInterface $cacheability): void {
+    $now = new DrupalDateTime('now');
+    $now->setTimezone(new \DateTimeZone(DateTimeItemInterface::STORAGE_TIMEZONE));
+    switch ($this->getConfiguration()['time']) {
+      case 'past':
+        $query->condition('oe_event_dates.end_value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), "<=");
```
sergepavle, line 62: From first looks, we have to use `<` not '<='. I could be wrong but maybe better to recheck.
So this question more related to the edge-case problem.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,114 @@
...
+    $future_event = Node::create($values);
+    $future_event->save();
+
+    /** @var \Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterPluginManager $plugin_manager */
+    $plugin_manager = $this->container->get('plugin.manager.oe_link_lists.internal_source_filter');
+    /** @var \Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterInterface $plugin */
+    $plugin = $plugin_manager->createInstance('event_link_source_filter', []);
+
+    // Plugin applies to events.
+    $this->assertTrue(TRUE, $plugin->isApplicable('node', 'oe_event'));
```
sergepavle, line 84: Probably it should be:
```suggestion
    $this->assertTrue($plugin->isApplicable('node', 'oe_event'));
```

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,114 @@
...
+    // Plugin does not apply to news.
+    $this->assertTrue(TRUE, $plugin->isApplicable('node', 'oe_news'));
```
sergepavle, line 86: I suppose it is should be:
```suggestion
    $this->assertFalse($plugin->isApplicable('node', 'oe_news'));
```

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,98 @@
...
+    return [
+      'time' => 'upcoming',
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isApplicable(string $entity_type, string $bundle): bool {
+    $applicable_entity_types = $this->pluginDefinition['entity_types'];
```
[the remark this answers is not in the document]
imanoleguskiza: The last array is not indexed with the bundle names so that would fail. I did however simplify it a bit, let me know what you think.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
imanoleguskiza: Agreed.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
imanoleguskiza: I think you are right, I removed the equal operators.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
imanoleguskiza: Fixed

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
imanoleguskiza: Fixed

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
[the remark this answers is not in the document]
sergepavle: yes. my bad

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,104 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
+use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterInterface;
+use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterPluginBase;
+
+/**
+ * Event link source filter class.
+ *
+ * @InternalLinkSourceFilter(
+ *   id = "event_link_source_filter",
```
upchuk, line 19: `oe_content_event_period`

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,104 @@
...
+ *   label = @Translation("Event Link Source Filter"),
```
upchuk, line 20: Event period link source filter

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,104 @@
...
+ *   description = @Translation("Filters for link lists using events as the internal source."),
```
upchuk, line 21: Filters events by the period they are in: past or upcoming.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,104 @@
...
+ *   entity_types = {
+ *     "node" = {
+ *       "oe_event",
+ *     },
+ *   },
+ * )
+ */
+class EventLinkSourceFilter extends InternalLinkSourceFilterPluginBase implements InternalLinkSourceFilterInterface {
```
upchuk, line 29: EventPeriodFilter

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventLinkSourceFilter.php
```diff
@@ -0,0 +1,104 @@
...
+        $query->sort('oe_event_dates.value', 'ASC');
+        break;
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form['time'] = [
```
upchuk, line 83: Let's call these period instead of time

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,115 @@
...
+  public static $modules = [
+    'oe_link_lists',
+    'oe_link_lists_internal_source',
+  ];
+
+  /**
+   * Test that the event filter works as intended.
+   */
+  public function testEventLinkSourceFilter(): void {
+
```
nagyad, line 30: Remove empty line.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -0,0 +1,104 @@
...
+use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
+use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\Core\Entity\Query\QueryInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
+use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterInterface;
+use Drupal\oe_link_lists_internal_source\InternalLinkSourceFilterPluginBase;
+
+/**
+ * Event link source filter class.
```
nagyad, line 16: Event period link source filter class.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -0,0 +1,104 @@
...
+  public function defaultConfiguration() {
+    return [
+      'period' => self::UPCOMING,
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isApplicable(string $entity_type, string $bundle): bool {
```
upchuk, line 53: Is this needed? i don't think so. it's done based on the annotation. Please see InternalLinkSourceFilterPluginManager which loads the applicable filters.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public static $modules = [
+    'oe_link_lists',
+    'oe_link_lists_internal_source',
+  ];
+
+  /**
+   * Test that the event filter works as intended.
+   */
+  public function testEventLinkSourceFilter(): void {
```
upchuk, line 29: Let's add https://www.drupal.org/project/datetime_testing and use this for testing properly the times. We are already using in the theme for the Behat extension.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -0,0 +1,104 @@
...
+      return TRUE;
+    }
+    return FALSE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function apply(QueryInterface $query, array $context, RefinableCacheableDependencyInterface $cacheability): void {
+    $now = new DrupalDateTime('now');
```
upchuk, line 64: We need to apply time caching here.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function apply(QueryInterface $query, array $context, RefinableCacheableDependencyInterface $cacheability): void {
+    $now = new DrupalDateTime();
+    $current_time = $this->time->getCurrentTime();
+    $now->setTimestamp($current_time);
+    $now->setTimezone(new \DateTimeZone(DateTimeItemInterface::STORAGE_TIMEZONE));
+    $cacheability->addCacheTags($this->timeBasedCacheTagGenerator->generateTags($now->getPhpDateTime()));
```
ademarco, line 109: Shouldn't we create caching tag based on the newest/oldest event node? Like that we are creating caching tags that are based on the first time a user visits a page where this is displayed.

imanoleguskiza: I'm not sure we need to test that caches are invalidated on cron, shouldn't that need to happen on oe_time_caching?

ademarco: Yes I meant that the oe_time_caching module will invalidate them on cron. But ok we can also not test it here as that's tested in the module already.

modules/oe_content_event/oe_content_event.info.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
   - field_group:field_group
   - inline_entity_form:inline_entity_form
   - entity_reference_revisions:entity_reference_revisions
+  - oe_time_caching:oe_time_caching
```
ademarco, line 22: This module is not a strictly a dependency as using that plugin is optional. We should add a section in the readme about integration with link list and there we say that we need this module too, in case we wish to use it.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -0,0 +1,176 @@
...
+
+      case self::UPCOMING:
+        $query->condition('oe_event_dates.value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), ">");
+        $query->sort('oe_event_dates.value', 'ASC');
+        // Time based cache tags need to be invalidated when the next
+        // upcoming event starts so execute the query and extract the
+        // starting date of the first event in the list.
+        $results = $query->execute();
+        $next_starting_event = $this->entityTypeManager->getStorage('node')->load(reset($results));
+        $next_starting_event_datetime = new DrupalDateTime($next_starting_event->oe_event_dates->value, new \DateTimeZone(DateTimeItemInterface::STORAGE_TIMEZONE));
```
ademarco, line 145: Let's not use the query we are altering to fetch results, I'm not sure if running execute() doesn't change any internal state for that object.

I'd create a separate method and use the Entity Query service for that, it might look as a duplication but it's more conceptually correct, I'd say.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -0,0 +1,191 @@
...
+        break;
+    }
+  }
+
+  /**
+   * Helper function to add time based cache tags to the query cacheability.
+   *
+   * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $cache
+   *   The refinable cacheability metadata for the current plugin.
+   * @param string $date_field_value_id
```
upchuk, line 147: Let's rename this variable and comment: 

```
@param string $date_value
  The date field column name: "value" for start date or "end_value" for end date.
```

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,167 @@
...
+    // Freeze the time at a specific point.
+    $static_time = new DrupalDateTime('2020-02-17 14:00:00', DateTimeItemInterface::STORAGE_TIMEZONE);
+    /** @var \Drupal\Component\Datetime\TimeInterface $datetime */
+    $time = $this->container->get('datetime.time');
+    $time->freezeTime();
+    $time->setTime($static_time->getTimestamp());
```
ademarco, line 39: We can set time one right before we run time related assertions, so the test becomes more readable.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,167 @@
...
+    $this->assertEqual($query_results, $past_events);
+    // Time based caches have been added and they correspond to the closest
+    // event end date.
+    $date_cache_tags = [
+      'oe_time_caching_date:2050',
+      'oe_time_caching_date:2050-05',
+      'oe_time_caching_date:2050-05-15',
+      'oe_time_caching_date:2050-05-15-12',
+    ];
+    $this->assertEqual($cache->getCacheTags(), $date_cache_tags);
```
ademarco, line 143: These assertions are exactly the same as the future one. We can have different generated tags by setting the current time between an event's start and end date? So we fall into this case:
```php
      case self::PAST:
        $query->condition('oe_event_dates.end_value', $now->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), "<");
        $query->sort('oe_event_dates.end_value', 'DESC');
```

imanoleguskiza: Actually we are already checking that that: the tags correspond to the event's end date (the day is different!)

ademarco: True that, my bad.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -0,0 +1,185 @@
...
+    // event end date.
+    $date_cache_tags = [
+      'oe_time_caching_date:2016',
+      'oe_time_caching_date:2016-05',
+      'oe_time_caching_date:2016-05-15',
+      'oe_time_caching_date:2016-05-15-12',
+    ];
+    $this->assertEqual($cache->getCacheTags(), $date_cache_tags);
+
+    // Configuring the filter to show all events shows all events..
```
ademarco, line 167: Double dot! :D

PR #190, opened 2020-04-28
OPENEUROPA-3058: Set event type widget to select.

nagyad:


oe_content.post_update.php
```diff
@@ -129,3 +130,24 @@ function oe_content_post_update_00004(): void {
...
   }
 }
+
+/**
+ * Change widget for "Event type" field to skos select in form display.
+ */
+function oe_content_post_update_00005() {
+  if (!\Drupal::moduleHandler()->moduleExists('oe_content_event')) {
+    return 'Module oe_content_event is not enabled.';
+  }
```
ademarco, line 140: Can't we just move this in `oe_content_event`? So it's ran only if that module is enabled?

composer.json
```diff
@@ -78,6 +78,9 @@
             "drupal/typed_link": {
                 "https://www.drupal.org/project/typed_link/issues/3085826": "https://www.drupal.org/files/issues/2019-10-04/typed_link-3085826-2.patch",
                 "https://www.drupal.org/project/typed_link/issues/3085817": "https://www.drupal.org/files/issues/2019-10-07/typed_link-3085817-3.patch"
+            },
+            "openeuropa/rdf_skos": {
+                "latest-master": "https://github.com/openeuropa/rdf_skos/compare/0.6.0...master.diff"
```
ademarco, line 83: Let's maybe release rdf_skos too as part of this ticket? So we don't have to do this.

oe_content.post_update.php
nagyad: Normally we prefer to place upgrade paths to the main module to ensure order. I thought I keep consistency.

composer.json
nagyad: Alright, I do it.

oe_content.post_update.php
ademarco: This code belongs to `oe_content_event`, which depends on `oe_content`, we have no order issues here.

modules/oe_content_event/oe_content_event.post_update.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Change widget for "Event type" field to skos select in form display.
+ */
+function oe_content_event_post_update_00001() {
+  $display = EntityFormDisplay::load('node.oe_event.default');
+  $content = $display->get('content');
+
+  if (empty($content['oe_event_type'])) {
+    return 'Event type field is not exposed on the form display.';
```
upchuk, line 20: Seeing that we are doing an update path here in the component, it means we are fixing something everywhere. So how come we are only fixing the default entity form display? What if there are more form displays that use the old select widget?

ademarco: We could also only do this on EWCMS, we have followed this rationale in other occasions too. @brummbar what do you think?

brummbar: This doesn't fix a bug indeed, so we should not have an opinion anymore on what's the best widget for the user that installed this module. Let's move this in EWPP.

PR #197, opened 2020-05-14
OPENEUROPA-3199: Fix Event scenario testing maxlength limits.

22Alexandra:


tests/features/event.feature
```diff
@@ -6,15 +6,19 @@ Feature: Event content creation
 
   @javascript
   Scenario: Length limited fields are truncating characters exceeding the configured limit.
-    Given I am logged in as a user with the "create oe_publication content, access content, edit own oe_publication content, view published skos concept entities" permission
-    When I visit "the Publication creation page"
+    Given I am logged in as a user with the "create oe_event content, access content, edit own oe_event content, view published skos concept entities" permission
```
imanoleguskiza, line 9: Same thing as the ewcms PR, in order to make sure we are truncating characters we have to check that the rest of the string is actually present

imanoleguskiza: I think I would move this test to the content-types.feature since its not something that applies only to events. I think ideally we would have an scenario outline and test this for all content types, but if you think its too much or too complicated we can let it go.

imanoleguskiza: Actually ignore the second comment, everything is tested in separate scenarios at the moment

22Alexandra: As long as the field is configured and the counter text is displayed I should not test the maxlength module's behavior.

PR #199, opened 2020-05-15
OPENEUROPA-3146: Create featured media field type.

nagyad:


modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
```diff
@@ -0,0 +1,127 @@
...
+ *   }
+ * )
+ */
+class FeaturedMediaFieldWidget extends EntityReferenceAutocompleteWidget {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+    $referenced_entities = $items->referencedEntities();
```
22Alexandra, line 30: No need to redo all the form, the parent class already does it. Do something like this:
`$element['featured_media'] = [`
`      '#type' => 'fieldset',`
  `    '#title' => $this->t('Featured media'),`
`    ] + parent::formElement($items, $delta, $element, $form, $form_state);`
  `  $element['featured_media']['caption'] = [`
`      '#title' => $this->t('Caption'),`
  `    '#type' => 'textarea',`
`      '#default_value' => $items[$delta]->caption,`
  `    '#rows' => 2,`
`    ];`

    return $element;`

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
```diff
@@ -0,0 +1,127 @@
...
+   * It provides useful information about the allowed media types to reference
+   * and a link to the media item list to create new media items.
+   *
+   * @param array $selection_settings
+   *   The entity reference field handler settings.
+   *
+   * @return mixed
+   *   The render array containing help texts.
+   */
+  protected function getMediaReferenceHelpText(array $selection_settings): array {
```
22Alexandra, line 94: This can be removed after changing the formElement.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -0,0 +1,88 @@
...
+ *     },
+ *     "caption" = {
+ *       "label" = @Translation("Caption"),
+ *       "translatable" = TRUE
+ *     },
+ *   },
+ *   list_class = "\Drupal\Core\Field\EntityReferenceFieldItemList"
+ * )
+ */
+class FeaturedMediaFieldItem extends EntityReferenceItem implements OptionsProviderInterface, PreconfiguredFieldUiOptionsInterface {
```
22Alexandra, line 38: It's enough to extend EntityReferenceItem.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -0,0 +1,88 @@
...
+
```
22Alexandra, line 39: Add isEmpty() method to check the caption field.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,198 @@
...
+      ->getFormDisplay($this->field->getTargetEntityTypeId(), $this->field->getTargetBundle())
+      ->setComponent($this->fieldStorage->getName(), $this->displayOptions);
+    $display->save();
+  }
+
+  /**
+   * Tests the featured media widget.
+   */
+  public function testFeaturedMediaWidget(): void {
+    // Visit the node add page.
```
22Alexandra, line 147: Test empty case.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -0,0 +1,88 @@
...
+  /**
```
22Alexandra, line 40: I can fill in Caption and save the node and when I edit it, my value is not saved.
We might need to add states if caption is filled in media field becomes required or a constraint.

22Alexandra:
The formatter doesn't render the media. It should have at least a simple formatter without ecl styling.

modules/oe_content_featured_media_field/config/schema/oe_content_featured_media_field.schema.yml
```diff
@@ -0,0 +1,41 @@
+field.field_settings.oe_featured_media:
```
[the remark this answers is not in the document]
22Alexandra: Screen Shot 2020-05-17 at 6 32 37 PM
I was wrong...it's needed. Sorry.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,198 @@
...
+    $assert_session->pageTextContains('See the media list (opens a new window) to help locate media.');
+    $assert_session->pageTextContains('Allowed media types: Image');
+
+    // Test the autocomplete functionality returns the image media.
+    $this->doAutocomplete($this->fieldStorage->getName());
+
+    $results = $page->findAll('css', '.ui-autocomplete li');
+
+    $this->assertCount(1, $results);
+    $assert_session->pageTextContains('Test image');
```
22Alexandra, line 182: The setUp method creates a node containing an image file but the values inside the field are not tested.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,198 @@
...
+        'sort' => [
+          'field' => '_none',
+        ],
+        'auto_create' => '0',
+      ],
+      'required' => FALSE,
+    ]);
+    $this->field->save();
+
+    // Create an image file.
```
22Alexandra, line 116: Test other media types not only images.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,227 @@
...
+        [
+          'target_id' => $image->id(),
+          'alt' => 'default alt',
+          'title' => 'default title',
+        ],
+      ],
+    ]);
+    $this->mediaEntity->save();
+
+    // Create a content type.
```
22Alexandra, line 124: Test other media types.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -0,0 +1,88 @@
...
+   * {@inheritdoc}
+   */
+  public function storageSettingsForm(array &$form, FormStateInterface $form_state, $has_data) {
+    $element = parent::storageSettingsForm($form, $form_state, $has_data);
+    // Don't allow any other entity type than media to be referenced.
+    $element['target_type']['#default_value'] = 'media';
+    $element['target_type']['#disabled'] = TRUE;
+
+    return $element;
+  }
```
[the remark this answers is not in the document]
nagyad: That's exactly what we do here.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldFormatter/FeaturedMediaFieldFormatter.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Plugin implementation of the 'Featured media as label' formatter.
+ *
+ * This is a default formatter for 'oe_featured_media' field type.
+ * In oe_theme the formatter won't be used, we take care of the rendering
+ * through preprocess hooks in a content type companion module.
+ *
+ * @FieldFormatter(
+ *   id = "oe_featured_media_label",
```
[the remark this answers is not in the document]
nagyad: Label indicates whats the formatter does. It's common to do this.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
nagyad: We have custom title, we have to make the help text work, and we need to place the field in a different element. In the parent method there is the autocreate entity we don't need as well.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
nagyad: Without this the help text won't work. Event if we would use use the parent.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
nagyad: We consider the field empty if there is no referenced entity. We don't need to look for the caption field.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
nagyad: It is tested below.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
nagyad: I will add states for this.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
nagyad: This is what we do in the Kernel test.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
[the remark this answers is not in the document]
22Alexandra: How is it useful to print a disabled element just to show the user that the reference type can be changed? The field type is called Featured media so it should reference only media. Feel free to leave it but it shouldn't be there in my opinion.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
22Alexandra: It's a field type with 2 items and we ignore one of them but I can save the node having values only in caption and then the value is not even saved.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
22Alexandra: But here you don't test the widget entirely. Fill in the fields and post the form then check the saved values inside the node.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
22Alexandra: Where? I don't see where you save the node with or without values and then asserting the saved values.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
nagyad: Plus when we use multi element field, it did not played out as I wanted due to the parent execution.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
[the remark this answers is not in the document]
nagyad: Done.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function schema(FieldStorageDefinitionInterface $field_definition) {
+    $schema = parent::schema($field_definition);
+
+    $schema['columns']['caption'] = [
+      'description' => 'The caption for the featured media.',
+      'type' => 'varchar_ascii',
```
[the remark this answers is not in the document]
nagyad: Done.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
22Alexandra: Ok then keep it as before.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
nagyad: We take care with #states to make the Media item required.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
nagyad: Here we are testing the behavior of the widget, not the field storage. That is tested in the Kernel test.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
nagyad: Done.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
nagyad: Done.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
nagyad: I added the test for asserting then ode.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -46,18 +43,6 @@ public static function defaultStorageSettings() {
...
-   */
-  public function storageSettingsForm(array &$form, FormStateInterface $form_state, $has_data) {
-    $element = parent::storageSettingsForm($form, $form_state, $has_data);
-    // Don't allow any other entity type than media to be referenced.
-    $element['target_type']['#default_value'] = 'media';
-    $element['target_type']['#disabled'] = TRUE;
-
-    return $element;
-  }
-
```
22Alexandra: I'm sorry if I wasn't clear. The storageSettingsForm method it is needed because without it we can change the reference type:
Screen Shot 2020-05-18 at 2 37 00 PM
but the function has to return an empty array as mentioned before, in order to be able to choose the cardinality:
Screen Shot 2020-05-18 at 2 39 33 PM

do smth like this:
```/**
   * {@inheritdoc}
   */
  public function storageSettingsForm(array &$form, FormStateInterface $form_state, $has_data) {
    return [];
  }```

Comment (22Alexandra) [reply modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php]: In any case, we need to decide if we reuse the storage under config/install or not to be able to have different settings for each entity.

**Comment by 22Alexandra** [comment]
Needs rebase.

File: modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php (hunk 3 above)
Comment (nagyad) [reply modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php]: Done.

File: modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php (hunk 14 above)
Comment (nagyad) [reply modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php]: We can see that later, for now it's there what is in the ticket.

File: modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php [hunk 15]
```diff
@@ -0,0 +1,236 @@
...
+    $assert_session->pageTextContains('Caption text');
+  }
+
+  /**
+   * Executes an autocomplete on a given field and waits for it to finish.
+   *
+   * @param string $field_name
+   *   The field name.
+   */
+  protected function doAutocomplete(string $field_name): void {
```
upchuk, line 229: What a missed opportunity to pass the actual value to search for in the autocomplete..

modules/oe_content_featured_media_field/config/install/field.storage.node.oe_featured_media_field.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - media
+    - node
+    - oe_content_featured_media_field
+id: node.oe_featured_media_field
```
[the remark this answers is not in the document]
nagyad: Yes, to reuse it

modules/oe_content_featured_media_field/config/install/field.storage.node.oe_featured_media_field.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - media
+    - node
```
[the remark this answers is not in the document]
nagyad: Every other filed storage we ship in config/install.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
```diff
@@ -0,0 +1,132 @@
...
+      ];
+    }
+
+    $all_bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo('media');
+    $allowed_bundles = !empty($selection_settings['target_bundles']) ? $selection_settings['target_bundles'] : [];
+    $bundle_labels = array_map(function ($bundle) use ($all_bundles) {
+      return $all_bundles[$bundle]['label'];
+    }, $allowed_bundles);
+
+    if (!empty($bundle_labels)) {
```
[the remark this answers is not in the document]
nagyad: Yes, when you create the field, these are not populated on that page.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,227 @@
...
+   * @var \Drupal\media\Entity\Media
+   */
+  protected $mediaEntity;
+
+  /**
+   * The entity view display used for test.
+   *
+   * @var \Drupal\Core\Entity\Display\EntityViewDisplayInterface
+   */
+  protected $display;
```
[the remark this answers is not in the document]
nagyad: We are using the display to build for rendering.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaFieldItem.php
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public static function schema(FieldStorageDefinitionInterface $field_definition) {
+    $schema = parent::schema($field_definition);
+
+    $schema['columns']['caption'] = [
+      'description' => 'The caption for the featured media.',
+      'type' => 'varchar',
+      'length' => 255,
```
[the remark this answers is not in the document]
nagyad: No, because previously the field was also 255.

modules/oe_content_featured_media_field/config/schema/oe_content_featured_media_field.schema.yml
```diff
@@ -0,0 +1,41 @@
...
+
+field.storage_settings.oe_featured_media:
+  type: mapping
+  label: 'Featured media field storage settings'
+  mapping:
+    target_type:
+      type: string
+      label: 'Type of media item to reference'
+
+field.value.oe_featured_media:
```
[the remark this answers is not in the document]
nagyad: I checked and entity reference and entity reference revisions, both add the target_id in the schema for default value.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaFieldWidget.php
```diff
@@ -0,0 +1,132 @@
...
+      // Entity reference field items are handling validation themselves via
+      // the 'ValidReference' constraint.
+      '#validate_reference' => FALSE,
+      '#maxlength' => 1024,
+      '#default_value' => isset($referenced_entities[$delta]) ? $referenced_entities[$delta] : NULL,
+      '#size' => $this->getSetting('size'),
+      '#placeholder' => $this->getSetting('placeholder'),
+      '#states' => [
+        'required' => [
+          ':input[name="' . $this->fieldDefinition->getName() . '[' . $delta . '][featured_media][caption]"]' => ['filled' => TRUE],
```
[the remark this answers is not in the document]
nagyad: No it doesn't. But it is true for core field as well. Do you have another idea to do this?

modules/oe_content_featured_media_field/config/install/field.storage.node.oe_featured_media_field.yml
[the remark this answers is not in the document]
upchuk: That makes absolutely no sense. Rename it

modules/oe_content_featured_media_field/src/Plugin/Field/FieldFormatter/FeaturedMediaFieldFormatter.php
```diff
@@ -0,0 +1,47 @@
...
+ *   label = @Translation("Label"),
+ *   description = @Translation("Display the label of the referenced media entity and the caption."),
+ *   field_types = {
+ *     "oe_featured_media"
+ *   }
+ * )
+ */
+class FeaturedMediaFieldFormatter extends EntityReferenceLabelFormatter {
```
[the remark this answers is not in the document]
upchuk: Screenshot 2020-05-20 09 13 27

Let's ensure some consistency in the names here:

```
FeaturedMediaFieldItem
FeaturedMediaAutocompleteWidget
FeaturedMediaLabelFormatter
```

modules/oe_content_featured_media_field/src/Plugin/Field/FieldFormatter/FeaturedMediaFieldFormatter.php
```diff
@@ -0,0 +1,47 @@
...
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $elements = [];
+
+    $parent_elements = parent::viewElements($items, $langcode);
+
+    foreach ($parent_elements as $delta => $parent_element) {
+      $elements[$delta]['featured_media'] = $parent_element;
+      // Simply add the caption as the next element after the media link.
+      $elements[]['caption'] = [
```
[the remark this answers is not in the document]
upchuk: what about this??

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaAutocompleteWidget.php
```diff
@@ -0,0 +1,163 @@
...
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+    $element['featured_media'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Featured media'),
+    ];
+
+    $element['featured_media'] += parent::formElement($items, $delta, $element, $form, $form_state);
+    // Unset title_display to ensure the element title is always visible.
```
upchuk, line 82: Why? The title display is configurable no? Why are we removing field API capability of a) choosing the title and b) showing/hiding it?

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaAutocompleteWidget.php
```diff
@@ -0,0 +1,163 @@
...
+      ],
+    ];
+
+    $element['featured_media']['caption'] = [
+      '#title' => $this->t('Caption'),
+      '#description' => $this->t('The caption that goes with the referenced media.'),
+      '#type' => 'textarea',
+      '#default_value' => $items[$delta]->caption,
+      '#rows' => 2,
+      '#weight' => $delta,
```
upchuk, line 98: What's this weight doing here? You are ordering the caption and autocomplete between themselves input using the field delta?

modules/oe_content_featured_media_field/config/schema/oe_content_featured_media_field.schema.yml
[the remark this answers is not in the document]
upchuk: Mm...its's probably wrong in core then..

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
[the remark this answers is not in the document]
upchuk: And why is it a class property? Pease read carefully my comments before replying.

modules/oe_content_featured_media_field/config/install/field.storage.node.oe_featured_media_field.yml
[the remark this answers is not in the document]
nagyad: I did rename in the first place.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldFormatter/FeaturedMediaFieldFormatter.php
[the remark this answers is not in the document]
nagyad: I did it, I don't know why it's not there. :(

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaAutocompleteWidget.php
nagyad: I agree with you, I assigned the parent title to the fieldset title. But the the title_display can't be set on form_display config for example and we want to make sure that the Media item title of the reference field is always visible. The entity reference field is using this flag when value is unlimited to hide the title. If I don't usnet it, it will hide it. So the underlined title in the screen shot will disappear.
image

modules/oe_content_featured_media_field/config/install/field.storage.node.oe_featured_media_field.yml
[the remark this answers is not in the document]
upchuk: OK, sorry

modules/oe_content_featured_media_field/src/Plugin/Field/FieldType/FeaturedMediaItem.php
```diff
@@ -50,6 +51,15 @@ public function storageSettingsForm(array &$form, FormStateInterface $form_state
     return [];
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public static function getPreconfiguredOptions(): array {
```
upchuk, line 57: Good job

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,226 @@
...
+    $media_entity->save();
+
+    // Create another media entity.
+    $media_entity = Media::create([
+      'bundle' => 'document',
+      'name' => 'Test media',
+    ]);
+    $media_entity->save();
+
+    $this->fieldStorage = FieldStorageConfig::create([
```
brummbar, line 94: I don't like too much the idea of using properties to store field instances and such, I think they belong to the test method itself most of the time. But talking about this case, it's clear we don't need to store them in properties. `$this->field` is not even used, while `$this->fieldStorage` is used once and to get its name (which we define once) so we can refer to it directly by its name `featured_media_field`.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,226 @@
...
+      'type' => 'oe_featured_media_label',
+      'label' => 'above',
+      'settings' => [
+        'link' => TRUE,
+      ],
+    ];
+
+    // Prepare the default form display for rendering.
+    $display = \Drupal::service('entity_display.repository')
+      ->getFormDisplay($this->field->getTargetEntityTypeId(), $this->field->getTargetBundle())
```
brummbar, line 144: `$this->field->getTargetEntityTypeId()` is node, and `$this->field->getTargetBundle()` is 'page'. We can just use those. Yet another reason for not storing the properties, we don't need them, they are hardcoded.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    $this->createContentType(['type' => 'page']);
+
+    $user = $this->drupalCreateUser([
```
brummbar, line 62: Let's move this login part to the other method, so we can put a comment on which user is logging like we did in line 174.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
brummbar: See \Drupal\Tests\field\Functional\Number\NumberFieldTest::testNumberDecimalField for reference.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,226 @@
...
+    ]));
+    $this->drupalGet('node/add/page');
+
+    // Assert the help texts with media overview permission.
+    $assert_session->pageTextContains('Start typing the name of the Media.');
+    $assert_session->pageTextContains('You can manage all the media items on this page.');
+    $assert_session->pageTextContains('Allowed media types: Image, Document');
+
+    // Test that the Media item field turns required once the Caption is filled.
+    $this->assertFalse($page->findField('Media item')->hasAttribute('required'));
```
brummbar, line 189: Let's not mix usages of assert_session and page. The method `$assert_session->fieldExists()` returns the dom element, let's stick to assert_session.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,226 @@
...
+    $page->fillField('Caption', 'Caption text');
+    $this->assertTrue($page->findField('Media item')->hasAttribute('required'));
+
+    // Test the autocomplete functionality returns the created media items.
+    $this->doAutocomplete($this->fieldStorage->getName(), 'Test');
```
brummbar, line 194: Here since we know the field name we will pass the string.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTest.php
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Executes an autocomplete on a given field and waits for it to finish.
+   *
+   * @param string $field_name
+   *   The field name.
+   * @param string $value
+   *   The value to look for.
+   */
+  protected function doAutocomplete(string $field_name, string $value): void {
```
brummbar, line 219: The field thought needs to be a featured media one, since we add `[0][featured_media][target_id]`. So the name is a bit misleading. We use it only once, we can merge it in the main method too. Or rename this better.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,240 @@
...
+          ],
+        ],
+        'sort' => [
+          'field' => '_none',
+        ],
+        'auto_create' => '0',
+      ],
+      'required' => FALSE,
+    ]);
+    $this->field->save();
```
brummbar, line 134: Oh almost all of the code of the field creation is the same as the previous test. Let's create a trait that allows to create such a field, getting a few parameters (entity type, bundle, field name, and such).

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,240 @@
...
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $entity_type_manager->load($node->id());
+
+    $expected = [
+      [
+        'target_id' => $this->mediaEntity->id(),
+        'caption' => 'Image caption text',
+      ],
+    ];
+    // Assert the base field values.
```
brummbar, line 167: Let's remove the word base, it's misleading indeed. Assert the field values should be enough. Also why we test the node label? Let's drop it.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,240 @@
...
+    // Create a node.
+    $node = Node::create($values);
+    $node->save();
+
+    $entity_type_manager = $this->container->get('entity_type.manager')->getStorage('node');
+    $entity_type_manager->resetCache();
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $entity_type_manager->load($node->id());
+
+    $expected = [
```
brummbar, line 161: `$expected === $values['featured_media_field']`.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,240 @@
...
+    // Test the rendering of the formatter with the test node.
+    // Prepare the default view display for rendering.
+    $display_options = [
+      'type' => 'oe_featured_media_label',
+      'label' => 'above',
+      'settings' => [
+        'link' => TRUE,
+      ],
+    ];
+    $display = $this->container->get('entity_display.repository')
```
brummbar, line 180: The rendering is tested in the browser test, why duplicate it here?

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
```diff
@@ -0,0 +1,240 @@
...
+          'caption' => 'Image caption text',
+        ],
+      ],
+    ];
+    // Create a node.
+    $node = Node::create($values);
+    $node->save();
+
+    // Assert the field values.
+    $this->assertEquals('My second node', $node->label());
```
brummbar, line 206: We should reload the node from the storage.

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
nagyad: Good idea. But I'm not sure because the field settings are slightly different in each test. In the end I would end up with a method where I need to feed most of the field attributes as argument.

brummbar: Oh in that case ignore my remark 😃

modules/oe_content_featured_media_field/tests/src/Kernel/FeaturedMediaFieldTest.php
nagyad: It's not true, because the target_id is defined here, but above in $values['featured_media_field'] it's not defined.

PR #205, opened 2020-05-25
OPENEUROPA-3216: Remove drupal-coder version.

nagyad:


composer.json
```diff
@@ -9,6 +9,7 @@
         "drupal/core": "^8.8",
         "drupal/linkit": "~5.0-beta8",
         "drupal/maxlength": "~1.0@beta",
+        "drupal/typed_link": "~1.0",
```
brummbar, line 12: This is not the correct solution. Typed link is not a dependency of oe_content anymore (see #165).
So the correct fix would actually be removing it from the info.yml of oe_content and some other fixes in the submodule.
Let's remove this part.

nagyad: Removed.

PR #209, opened 2020-06-04
OPENEUROPA-3235: Featured media entity browser widget.

22Alexandra:


modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,269 @@
...
+   * @SuppressWarnings(PHPMD.NPathComplexity)
+   */
+  protected function formElementItemsEntities(FieldItemListInterface $items, int $delta, array $element, FormStateInterface $form_state) {
+    $entities = [];
+    $entity_type = $this->fieldDefinition->getFieldStorageDefinition()->getSetting('target_type');
+    $entity_storage = $this->entityTypeManager->getStorage($entity_type);
+
+    // Find IDs from target_id element (it stores selected entities in form).
+    // This was added to help solve a really edge casey bug in IEF.
+    if (($target_id_entities = $this->getEntitiesByTargetId($element, $form_state)) !== FALSE) {
```
upchuk, line 141: Seems to work in our case also without this in IEF.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,269 @@
...
+      if ($last_parent === 'add_more') {
+        $parents = $triggering_element['#parents'];
+        // Remove the button key.
+        array_pop($parents);
+        $value = $form_state->getValue($parents);
+        return EntityBrowserElement::processEntityIds($value[$delta]['target_id']);
+      }
+    };
+
+    // We are loading for for the first time so we need to load any existing
```
upchuk, line 199: for for

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,269 @@
...
+      $entity = $entity_storage->load($items[$delta]->target_id);
+      if (!empty($entity)) {
+        $entities[] = $entity;
+      }
+    }
+    return $entities;
+  }
+
+  /**
+   * Overrides \Drupal\Core\Field\WidgetBase::formMultipleElements().
```
upchuk, line 212: Just {@inheritdoc} here

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,269 @@
...
+   */
+  protected function formMultipleElements(FieldItemListInterface $items, array &$form, FormStateInterface $form_state) {
+    $elements = parent::formMultipleElements($items, $form, $form_state);
+
+    $field_name = $this->fieldDefinition->getName();
+    $cardinality = $this->fieldDefinition->getFieldStorageDefinition()->getCardinality();
+    $parents = $form['#parents'];
+
+    // Determine the number of possible items.
+    switch ($cardinality) {
```
upchuk, line 225: Why you complicating this so much?

1. There is a #max_delta value in the $elements array that tells you the max delta
2. Anyway, it doesn't matter. You can just loop through Element::children() of elements and for each that has a "_weight", we cut the access no?

modules/oe_content_featured_media_field/tests/modules/oe_content_featured_media_test/oe_content_featured_media_test.info.yml
```diff
@@ -0,0 +1,11 @@
+name: OpenEuropa Content Featured Media Test
+description: Test module for Featured Media entity browser widget.
```
upchuk, line 2: Well, the test module is not JUST for the entity browser widget no?

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,206 @@
...
+    $this->getSession()->getPage()->pressButton('Save');
+
+    // Assert that the values were saved.
+    $this->assertSession()->pageTextContains('Featured media field');
+    $this->assertSession()->pageTextContains('Image 1');
+    $this->assertSession()->pageTextContains('Image 2');
+    $this->assertSession()->pageTextContains('Image 1 caption');
+    $this->assertSession()->pageTextContains('Image 2 caption');
+
+    // Edit the node to remove the first media.
```
upchuk, line 197: Before removing, reorder the fields and see if they got reordered (you can load the node and assert correct deltas). See `TableDragTest` for way to table drag.

Then remove one, including the caption ,and assert it got removed completely also after save.

modules/oe_content_featured_media_field/tests/modules/oe_content_featured_media_test/oe_content_featured_media_test.info.yml
22Alexandra: For now, it is since it only contains a view and and entity browser. I can update the description.

modules/oe_content_featured_media_field/README.md
```diff
@@ -0,0 +1,15 @@
+# OpenEuropa Content Featured Media Field
+
+This module provides the Featured media field type containing a media reference field and a caption textfield.
+It includes an autocomplete widget, and an entity browser widget for the field type.
```
upchuk, line 4: No comma between widget and and.

modules/oe_content_featured_media_field/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Installation
+
+Before enabling this module, make sure that the following module is present in your codebase by adding it to your
```
upchuk, line 8: The dependency is only needed if you want to use the entity browser based widget.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,246 @@
...
+   * {@inheritdoc}
+   */
+  protected function formMultipleElements(FieldItemListInterface $items, array &$form, FormStateInterface $form_state) {
+    $elements = parent::formMultipleElements($items, $form, $form_state);
+
+    // Do not allow access to the weight select in order to avoid mixed data
+    // between media items and captions.
+    foreach (Element::children($elements) as $key) {
+      if (isset($elements[$key]['_weight'])) {
+        $elements[$key]['_weight']['#type'] = 'hidden';
```
upchuk, line 215: Let's add a comment here as to why we use the `hidden` element and not just disable it.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    // Check that 'Image 1' media item is placed before 'Image 2'.
+    $this->assertOrderInPage(['Image 1', 'Image 2']);
+
+    // Fill in the other fields and save the node.
+    $this->getSession()->getPage()->fillField('featured_media_field[0][caption]', 'Image 1 caption');
+    $this->getSession()->getPage()->fillField('featured_media_field[1][caption]', 'Image 2 caption');
+    $this->getSession()->getPage()->fillField('Title', 'Test entity browser widget');
+    $this->getSession()->getPage()->pressButton('Save');
+
+    // Assert that the values were saved.
```
upchuk, line 194: Well, if there was an error, nothing will happen and it will be a saved positive. Assert the save message is printed on the page or reload.

Load also the node and assert the values.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $this->assertSession()->pageTextContains('Image 1');
+    $this->assertSession()->pageTextContains('Image 2');
+    $this->assertSession()->pageTextContains('Image 1 caption');
+    $this->assertSession()->pageTextContains('Image 2 caption');
+    // Assert items order is the same as it was in the form.
+    $this->assertOrderInPage(['Image 1', 'Image 2']);
+
+    // Edit the node to reorder field items.
+    $this->drupalGet('node/1/edit');
+    file_put_contents('test.html', $this->getSession()->getPage()->getOuterHtml());
```
upchuk, line 205: Let's not put file contents :P

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $dragged = $this->getSession()->getPage()->find('css', 'table#featured-media-field-values > tbody > tr:nth-child(1) a.tabledrag-handle');
```
upchuk, line 206: let's call this `$handle`

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $target = $this->getSession()->getPage()->find('css', 'table#featured-media-field-values > tbody > tr:nth-child(2) a.tabledrag-handle');
+    $dragged->dragTo($target);
+    $this->assertSession()->assertWaitOnAjaxRequest();
+
+    // Check that 'Image 1' media item is placed after 'Image 2'.
+    $this->assertOrderInPage(['Image 2', 'Image 1']);
+    $this->getSession()->getPage()->pressButton('Save');
+
+    // Assert items order was saved.
```
upchuk, line 215: Again, assert the save message to be sure there was no error. And after, load the node and assert correct values in storage.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $this->assertOrderInPage(['Image 2', 'Image 1']);
+
+    // Edit the node to remove the first item.
+    $this->drupalGet('node/1/edit');
+    $this->getSession()->getPage()->pressButton('Remove');
```
upchuk, line 220: Please specify an exact Remove button.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    // Assert the image was removed from the field.
+    $this->assertSession()->pageTextNotContains('Image 2');
+    $this->assertSession()->buttonExists('Select images');
+    $this->getSession()->getPage()->pressButton('Save');
```
upchuk, line 225: Assert save message and then load and assert values as well.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+
+    // Assert that the update was saved.
+    $this->assertSession()->pageTextContains('Featured media field');
```
upchuk, line 228: And where is the assertion that the removal worked?

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $this->assertSession()->pageTextContains('Image 1');
+    $this->assertSession()->pageTextContains('Image 1 caption');
+  }
+
+  /**
+   * Asserts items are placed in markup in the given order.
+   */
+  protected function assertOrderInPage(array $items) {
```
upchuk, line 236: return type

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $session = $this->getSession();
+    $text = $session->getPage()->getHtml();
+    $strings = [];
+    foreach ($items as $item) {
+      if (($pos = strpos($text, $item)) === FALSE) {
+        throw new ExpectationException("Cannot find '$item' in the page", $session->getDriver());
```
upchuk, line 242: You can simply assert that pageTextContains each of the items and don't have to throw this exception

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
upchuk, line 236: let's call this `$expected_strings`

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    $this->assertSession()->pageTextContains('Image 1 caption');
+  }
+
+  /**
+   * Asserts items are placed in markup in the given order.
+   */
+  protected function assertOrderInPage(array $items) {
+    $session = $this->getSession();
+    $text = $session->getPage()->getHtml();
+    $strings = [];
```
upchuk, line 239: let's call this `$actual_strings`

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,253 @@
...
+    foreach ($items as $item) {
+      if (($pos = strpos($text, $item)) === FALSE) {
+        throw new ExpectationException("Cannot find '$item' in the page", $session->getDriver());
+      }
+      $strings[$pos] = $item;
+    }
+    ksort($strings);
+    $ordered = implode(', ', array_map(function ($item) {
```
upchuk, line 247: What? I think you meant to order the actual found strings and not the passed ones no? I don't think this is even needed. You can assert the same and implode the actual found strings no?

modules/oe_content_featured_media_field/README.md
22Alexandra: image
😁

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
22Alexandra: I'm saving the $item in the $strings array on the position it was found in the markup.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
22Alexandra: $ordered just makes one string from the given items to be returned when the order is not the same. The message returned is indeed wrong.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
22Alexandra: Why $handle? This is the item that has to be dragged.

modules/oe_content_featured_media_field/README.md
upchuk: It does not separate two clauses. It's an enumeration: it includes and autocomplete widget and an entity browser. 

Mary ate an apple, and a pear ?!?!!?

ademarco:
When a field has the widget set and I visit its settings, then i get this error:

image

And the widget on the field setting page looks like this:

ademarco:
Also the caption should never be considered as required, even if the field is required:

image

The other widget only seems to mark it as required for existing fields, so there we might also have an issue:

image

upchuk:
✅  Validation to prevent submit of caption without media
✅  Not making the captions mandartory

The errors I cannot reproduce. Can you try again and let me know if you still have issues?

ademarco:
The Remove button seems not to work in the following scenario:

1. You have a multiple featured media field
2. You add one item to the field and save the node
3. You edit the node, add the same media to the second item and, without saving, you try to remove the media on the first item

Peek 2020-07-28 16-14

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -0,0 +1,254 @@
...
+        'link' => TRUE,
+      ],
+    ];
+
+    FieldStorageConfig::load('node.featured_media_field')
+      ->setCardinality(FieldStorageDefinitionInterface::CARDINALITY_UNLIMITED)
+      ->save();
+
+    /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+    $form_display = $this->container->get('entity_type.manager')
```
nagyad, line 47: Below you are accessing the service via "\Drupal::service('entity_display.repository')". Do we need to use container here?

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,385 @@
...
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface[]
+   *   The list of entities for the form element.
+   *
+   * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+   * @SuppressWarnings(PHPMD.NPathComplexity)
+   */
+  protected function formElementItemsEntities(FieldItemListInterface $items, int $delta, array $element, FormStateInterface $form_state) {
```
nagyad, line 260: Missing return type hint.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,385 @@
...
+ *   }
+ * )
+ */
+class FeaturedMediaEntityBrowserWidget extends EntityReferenceBrowserWidget {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    // Override some default settings that we don't want changed by the user.
```
nagyad, line 34: ... want to be changed ...

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,385 @@
...
+   * Caption field validator.
+   *
+   * @param array $element
+   *   The element.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   * @param array $complete_form
+   *   The entire form.
+   */
+  public function validateCaptionField(array &$element, FormStateInterface $form_state, array &$complete_form): void {
```
nagyad, line 222: $complete_form is not used in the function.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -0,0 +1,385 @@
...
+      // few levels up to read value of "target_id" element.
+      elseif ($triggering_element['#type'] == 'submit' && strpos($triggering_element['#name'], $this->fieldDefinition->getName() . '_remove_') === 0) {
+        $parents = array_merge(array_slice($triggering_element['#parents'], 0, -static::$deleteDepth), ['target_id']);
+      }
+
+      // Since we are using a delta, replace the second value after the field
+      // name key with the current delta being requested.
+      $parents[$field_name_key + 1] = $delta;
+
+      if (isset($parents) && $value = $form_state->getValue($parents)) {
```
nagyad, line 304: I think parents will be always set.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTestBase.php
```diff
@@ -0,0 +1,108 @@
...
+    'media',
+    'media_test_source',
+    'oe_media',
+    'oe_content_featured_media_field',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $defaultTheme = 'stark';
```
nagyad, line 37: Why do need stark instead of bartik?

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTestBase.php
```diff
@@ -0,0 +1,108 @@
...
+
+    $this->createContentType(['type' => 'page']);
+
+    // Create an image file.
+    \Drupal::service('file_system')->copy($this->root . '/core/misc/druplicon.png', 'public://example.jpg');
+    $image = File::create(['uri' => 'public://example.jpg']);
+    $image->save();
+
+    // Create 2 image media entities.
+    $media_entity = Media::create([
```
nagyad, line 53: We are not really using $media_entity for anything in this method, we could eliminate it completely and chain the ->save() after the create().

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTestBase.php
```diff
@@ -0,0 +1,108 @@
...
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    $this->createContentType(['type' => 'page']);
+
+    // Create an image file.
+    \Drupal::service('file_system')->copy($this->root . '/core/misc/druplicon.png', 'public://example.jpg');
+    $image = File::create(['uri' => 'public://example.jpg']);
```
nagyad, line 49: We are not really using $image for anything in this method, we could eliminate it completely and chain the ->save() after the create(). Just like below with FieldStorageConfig, for consistency.

nagyad:
Please add the module in the runner to be enabled.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
upchuk: no

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
upchuk: so? it's passed from the validator.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTestBase.php
upchuk: you never test with bartik

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaFieldWidgetTestBase.php
upchuk: it's fine..

PR #214, opened 2020-06-15
OPENEUROPA-3299: News featured media to reference video iframes as well.

upchuk:


upchuk:
Closing as this is only for EWCMS.

PR #222, opened 2020-07-06
OPENEUROPA-3048: Use sparql storage.

imanoleguskiza:
### Description

Use sparql storage.
### Change log

- Added:
- Changed: Use sparql storage.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/Kernel/ShortTitleTokenReplaceTest.php
```diff
@@ -37,10 +38,11 @@ class ShortTitleTokenReplaceTest extends RdfKernelTestBase {
   /**
    * {@inheritdoc}
    */
-  protected function setUp() {
+  protected function setUp(): void {
```
sergepavle, line 41: Should we specify return type in overridden method?

imanoleguskiza: I think do yes, phpstorm already complains about it otherwise.

sergepavle (dismissed):

tests/Kernel/ShortTitleTokenReplaceTest.php
nagyad: We can specify the same, we can't change though.

nagyad (dismissed):

brummbar (dismissed):

brummbar (dismissed):

brummbar (approved):

ademarco (approved):

PR #223, opened 2020-07-06
EWPP-31: Make Departments a shared field.

julien-:


modules/oe_content_project/config/install/core.entity_view_display.node.oe_project.default.yml
```diff
@@ -47,6 +47,14 @@ content:
     settings:
       link: true
     third_party_settings: {  }
+  oe_departments:
+    type: skos_concept_entity_reference_label
+    weight: 4
+    region: content
+    label: above
+    settings:
+      link: true
```
nagyad, line 56: I think this should not be a clickable link, right?

julien-: @nagyad , it seems to be clickable
https://webgate.ec.europa.eu/fpfis/wikis/download/attachments/388531445/Project%20CT%20-%20Desktop.png?version=3&modificationDate=1585157193125&api=v2

drishu: you have a good eye :)

nagyad: The mockup is wrong. The link would lead to the skos entity, that is not a page to view for an end user. So unless there is no dedicated logic where this should land, this should not be clickable.

julien-: Fixed

PR #225, opened 2020-07-08
EWPP-26: Extend events filter for ongoing events.

nagyad:


modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
```diff
@@ -33,6 +33,11 @@
  */
 class EventPeriodFilter extends InternalLinkSourceFilterPluginBase implements InternalLinkSourceFilterInterface, ContainerFactoryPluginInterface {
 
+  /**
+   * Option for ongoing events.
+   */
+  const ONGOING = 2;
```
sergepavle, line 39: Maybe even nicer to use 0 integer value for ongoing event constant (as it is a moment between past and upcoming).

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -164,6 +176,29 @@ public function testEventLinkSourceFilter(): void {
     ];
     $this->assertEqual($cache->getCacheTags(), $date_cache_tags);
 
+    // Set the time to test ongoing events.
```
sergepavle, line 179: Looks like we are missing tests of edge cases when the current time is '2016-05-10T12:00:00' or '2050-05-15T12:00:00'. As I understand we are expecting at this moment anyway to have the ongoing event.

modules/oe_content_event/src/Plugin/InternalLinkSourceFilter/EventPeriodFilter.php
nagyad: You can't use 0 there, it will considered as empty and it will show all events.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
nagyad: Done.

nagyad:
Rebased.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -110,15 +122,16 @@ public function testEventLinkSourceFilter(): void {
     $future_events = [
       $upcoming_event->id() => $upcoming_event->id(),
       $future_event->id() => $future_event->id(),
+      $ongoing_event->id() => $ongoing_event->id(),
     ];
     $this->assertEqual($query_results, $future_events);
```
imanoleguskiza, line 127: Actually after introducing the ongoing event the two arrays are not equal: the order is different since the ongoing event should be the first.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -164,6 +177,31 @@ public function testEventLinkSourceFilter(): void {
...
+    // Set the time to the end date of an ongoing event.
+    $static_time = new DrupalDateTime('2050-05-15 12:00:00', DateTimeItemInterface::STORAGE_TIMEZONE);
+    $time->setTime($static_time->getTimestamp());
+    $query = $storage->getQuery();
+    // Configuring the filter for upcoming events.
+    $plugin->setConfiguration(['period' => EventPeriodFilter::UPCOMING]);
+    $cache = new CacheableMetadata();
+    $plugin->apply($query, [], $cache);
+    $query_results = $query->execute();
+    $ongoing_events = [
```
imanoleguskiza, line 189: This test makes no sense, according to the comment we "Set the time to the end date of an ongoing event." (which by the way is also the end date for the upcoming event, which makes things even more confusing), so when we use the "upcoming" filter I would assume that the ongoing event SHOULDN'T be in the results, and still it is (together with the upcoming event which has also ended by now).

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
```diff
@@ -164,6 +177,31 @@ public function testEventLinkSourceFilter(): void {
...
+      $ongoing_event->id() => $ongoing_event->id(),
+      $upcoming_event->id() => $upcoming_event->id(),
+      $future_event->id() => $future_event->id(),
+    ];
+    $this->assertEqual($query_results, $ongoing_events);
+    // Time based caches have been added and they correspond to the closest
+    // event end date.
+    $date_cache_tags = [
```
imanoleguskiza, line 197: This also highlights the problem described above, now the cache won't be cleared until 2200 so this means that unless we clear it manually, the ongoing and upcoming events (which have finished) will be there for 150 years :(

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
nagyad: Yeah, but that is the requirement... :(

nagyad: We have identified a potential conflict with the fact that we are using the cache tags with the exact end date, but actually the end date is still the date of the event that is ongoing and the granularity of the cache is hour based. Ex.: End date today at 12:00:00 - at this exact time the event is still ongoing the event will become past event when the time hits 12:00:01, but the cache gets invalidated before and that causes inconsistencies. We have to discuss this problem with KUGLER Tamas to understand what should be the minimum granularity of the time caching.

modules/oe_content_event/tests/src/Kernel/EventLinkSourceFilterTest.php
nagyad: > We have identified a potential conflict with the fact that we are using the cache tags with the exact end date, but actually the end date is still the date of the event that is ongoing and the granularity of the cache is hour based. Ex.: End date today at 12:00:00 - at this exact time the event is still ongoing the event will become past event when the time hits 12:00:01, but the cache gets invalidated before and that causes inconsistencies. We have to discuss this problem with KUGLER Tamas to understand what should be the minimum granularity of the time caching.

PR #231, opened 2020-07-22
EWPP-33: Add Documents shared field on Project content type

drishu:
# EWPP-33

### Description

Add module scafold, and oe_document to project.

### Change log

- Added: oe_content_document_field module
- Changed:
   - updated project behat test
  - updated project context with new field


modules/oe_content_project/config/install/field.field.node.oe_project.oe_documents.yml
```diff
@@ -0,0 +1,32 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.node.oe_documents
+    - media.type.document
+    - node.type.oe_project
+  module:
+    - composite_reference
```
imanoleguskiza, line 9: This feels like it should have a dependency on oe_content_documents_field. Was this file generated through commands?

tests/features/project.feature
```diff
@@ -16,6 +16,7 @@ Feature: Project content creation
     And the following document:
       | name          | file       |
       | My Document 1 | sample.pdf |
+      | My Document 2 | sample.pdf |
```
imanoleguskiza, line 19: We should probably have separate file names for this so we can differentiate them after.

imanoleguskiza (changes requested):
The problem with this PR is that we are moving a configuration object from a module to another so if we enable the new oe_content_documents_field on ewcms we will get an error about the configuration already existing.

modules/oe_content_project/config/install/field.field.node.oe_project.oe_documents.yml
drishu: yes, it was exported, this https://github.com/openeuropa/oe_content/blob/74b5f8dd7cde1f56fc32963525e565294101492d/modules/oe_content_project/config/install/field.field.node.oe_project.oe_documents.yml#L5 is how drupal exported the dependency, and in the info.yml we have the ref to the module

tests/features/project.feature
drishu: We have dedicated regions https://github.com/openeuropa/oe_content/blob/74b5f8dd7cde1f56fc32963525e565294101492d/behat.yml.dist#L69 which ensure that the values are correctly assigned, would you agree that, taking above into consideration, its better to reuse existing files then adding more to the fixtures ?

imanoleguskiza: The problem is that because we check for "sample.pdf" in both cases, with this test we can never be sure about which of the two media items is being rendered. I don't think adding another fixture would make much of an impact? What to you think @ademarco ?

nagyad (changes requested):
Needs rebase.

modules/oe_content_project/config/install/field.field.node.oe_project.oe_documents.yml
nagyad: It's normal since the module adds a third-party setting to every field config.

nagyad (changes requested):

nagyad (changes requested):

modules/oe_content_publication/config/install/field.field.node.oe_publication.oe_documents.yml
```diff
@@ -2,7 +2,6 @@ langcode: en
 status: true
 dependencies:
   config:
-    - field.storage.node.oe_documents
```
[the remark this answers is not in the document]
drishu: If I add the module as dependency here then I get
`Configuration objects provided by <em class="placeholder">oe_content_publication</em> have unmet dependencies: <em class="placeholder">field.field.no  
  de.oe_publication.oe_documents (oe_content_documents_field)</em> ` dono why,
**but** the oe_content_documents_field module is already dependency of the publication module so we are good.

nagyad (dismissed):

tests/features/project.feature
ademarco: It would be better to use a different file, even if we have a dedicated region, as we aren't technically sure we are rendering the document "My Document 2" (that's what @imanoleguskiza is trying to say, I think).

Simply add another sample file to our fixture directory, commit it and use that one.

ademarco (changes requested):

modules/oe_content_documents_field/oe_content_documents_field.install
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Implements hook_install().
+ */
+function oe_content_documents_field_install() {
+  // If we are installing from config, we bail out.
+  if (\Drupal::isConfigSyncing() === TRUE) {
+    return;
+  }
+
+  // Does the field storage already exist ?
```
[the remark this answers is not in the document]
drishu: done

imanoleguskiza (dismissed):

modules/oe_content_documents_field/README.md
```diff
@@ -0,0 +1,4 @@
+# OpenEuropa Content Documents Field
+
+This module provides the Documents (oe_document) Field storage field, by enabling it you will be able to add Documents
```
ademarco, line 3: The field is called `oe_documents`.

tests/Behat/Content/Node/ProjectContentContext.php
```diff
@@ -47,6 +47,7 @@ public function alterProjectFields(BeforeParseEntityFieldsScope $scope): void {
       'Project period end date' => 'oe_project_dates:end_value',
       'Published' => 'status',
       'Reference' => 'oe_reference',
+      'Documents' => 'oe_document',
```
ademarco, line 50: Isn't the field call `oe_documents`?

ademarco (changes requested):

ademarco (approved):

imanoleguskiza (approved):

PR #232, opened 2020-07-22
EWPP-58: Add 'oe_reference' field in shared fields.

julien-:


julien-:
> You also need https://github.com/openeuropa/oe_content/blob/EPIC-Project/oe_content.post_update.php#L137
> And a PR which removes all this dev from EPIC-project which will be red in drone, but after this is merged you rebase and it will be green.

File added and [PR for EPIC-Project done](https://github.com/openeuropa/oe_content/pull/238)

config/install/field.storage.node.oe_reference.yml
```diff
@@ -0,0 +1,20 @@
...
+entity_type: node
+type: string
+settings:
+  max_length: 150
+  is_ascii: false
+  case_sensitive: false
+module: core
+locked: false
+cardinality: 1
+translatable: true
```
nagyad, line 17: This should not be translatable.

config/post_updates/00005_add_oe_reference_storage_field/field.storage.node.oe_reference.yml
```diff
@@ -0,0 +1,20 @@
...
+entity_type: node
+type: string
+settings:
+  max_length: 150
+  is_ascii: false
+  case_sensitive: false
+module: core
+locked: false
+cardinality: 1
+translatable: true
```
nagyad, line 17: This should not be translatable.

oe_content.info.yml
```diff
@@ -16,6 +16,7 @@ config_devel:
   install:
     - field.storage.node.oe_author
     - field.storage.node.oe_publication_date
+    - field.storage.node.oe_reference
```
ademarco, line 19: This is wrong right? the reference field shouldn't be exported on the main module.

modules/oe_content_reference/README.md
```diff
@@ -0,0 +1,9 @@
+# OpenEuropa Reference Field
```
ademarco, line 1:
```suggestion
# OpenEuropa Content Reference Code Field
```

modules/oe_content_reference/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This module provides the Reference Field storage field, by enabling it you will be able to add reference
+fields to your content types. Reference fields will be automatically styled when using OpenEuropa Theme.
```
ademarco, line 4: Let's use the description below, so we don't confuse users, as they might thing it has something to do with referencing entities.

> This module provides the storage configuration for a reference code field. A reference code field is a simple text field that can be used to store codes, IDs and other reference-like information. This field is used on several corporate content types, so it is packaged in its own, independent, module.

modules/oe_content_reference/oe_content_reference.info.yml
```diff
@@ -0,0 +1,13 @@
+name: OpenEuropa Content Reference Field
```
ademarco, line 1: Rename this module in `oe_content_reference_code_field`, so to avoid confusion with the other meaning of the word `reference` in Drupal.

modules/oe_content_reference/config/install/field.storage.node.oe_reference.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: node.oe_reference
```
ademarco, line 6: Rename this field into `oe_reference_code`, so to avoid confusion with the other meaning of the word reference in Drupal.

modules/oe_content_reference/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Usage
+
+By enabling this module you just get the possibility to add a 150 length, non translatable text field named `oe_reference`
```
ademarco, line 8:
```suggestion
By enabling this module you just get the possibility to add a 150 length, non translatable text field named `oe_reference_code`
```

modules/oe_content_reference/README.md
drishu: done

modules/oe_content_reference/README.md
drishu: done

modules/oe_content_reference/oe_content_reference.info.yml
drishu: done

modules/oe_content_reference/config/install/field.storage.node.oe_reference.yml
drishu: done

modules/oe_content_reference/README.md
drishu: done

PR #239, opened 2020-07-31
EWPP-94: Refactor the PURL module to allow for other entity types

imanoleguskiza:
### Description

Refactor the PURL module to allow for other entity types

### Change log

- Added: Event to resolve URLs.
- Changed: Moved supported entity types to configuration.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
+++ b/modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
@@ -1,2 +1,4 @@
 langcode: en
 base_url: 'https://data.ec.europa.eu/ewp/'
+supported_entity_types:
+  node: node
--- a/modules/oe_content_persistent/oe_content_persistent.post_update.php
+++ b/modules/oe_content_persistent/oe_content_persistent.post_update.php
@@ -11,7 +11,12 @@ declare(strict_types = 1);
  * Update PURL configuration.
  */
 function oe_content_persistent_post_update_00001(): void {
+  $supported_entity_types = \Drupal::getContainer()->getParameter('oe_content_persistent.supported_entity_types');
+  if (empty($supported_entity_types)) {
+    $supported_entity_types = ['node'];
+  }
+  $supported_entity_types = array_combine($supported_entity_types, $supported_entity_types);
   $purl_config = \Drupal::configFactory()->getEditable('oe_content_persistent.settings');
-  $purl_config->set('supported_entity_types', ['node' => 'node']);
+  $purl_config->set('supported_entity_types', $supported_entity_types);
   $purl_config->save();
 }
--- a/modules/oe_content_persistent/oe_content_persistent.post_update.php
+++ b/modules/oe_content_persistent/oe_content_persistent.post_update.php
@@ -11,7 +11,9 @@ declare(strict_types = 1);
  * Update PURL configuration.
  */
 function oe_content_persistent_post_update_00001(): void {
-  $supported_entity_types = \Drupal::getContainer()->getParameter('oe_content_persistent.supported_entity_types');
+  if (\Drupal::getContainer()->hasParameter('oe_content_persistent.supported_entity_types')) {
+    $supported_entity_types = \Drupal::getContainer()->getParameter('oe_content_persistent.supported_entity_types');
+  }
   if (empty($supported_entity_types)) {
     $supported_entity_types = ['node'];
   }
--- a/modules/oe_content_persistent/oe_content_persistent.post_update.php
+++ b/modules/oe_content_persistent/oe_content_persistent.post_update.php
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * Post update functions for OpenEuropa Content module.
+ * Post update functions for OpenEuropa Content Persistent module.
  */
 
 declare(strict_types = 1);
--- a/modules/oe_content_persistent/oe_content_persistent.post_update.php
+++ b/modules/oe_content_persistent/oe_content_persistent.post_update.php
@@ -1,0 +1,17 @@
+<?php
+
+/**
+ * @file
+ * Post update functions for OpenEuropa Content module.
+ */
+
+declare(strict_types = 1);
+
+/**
+ * Update PURL configuration.
+ */
+function oe_content_persistent_post_update_00001(): void {
+  $purl_config = \Drupal::configFactory()->getEditable('oe_content_persistent.settings');
+  $purl_config->set('supported_entity_types', ['node' => 'node']);
+  $purl_config->save();
+}
--- a/modules/oe_content_persistent/src/Controller/PersistentUrlController.php
+++ b/modules/oe_content_persistent/src/Controller/PersistentUrlController.php
@@ -6,9 +6,11 @@ namespace Drupal\oe_content_persistent\Controller;
 
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
-use Drupal\Core\Entity\TranslatableInterface;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\oe_content_persistent\ContentUuidResolverInterface;
+use Drupal\oe_content_persistent\Event\PersistentUrlResolverEvent;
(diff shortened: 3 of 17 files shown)


modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
```diff
@@ -1,2 +1,4 @@
 langcode: en
 base_url: 'https://data.ec.europa.eu/ewp/'
+supported_entity_types:
```
upchuk, line 3: Without an update path this is not BC..

modules/oe_content_persistent/src/Controller/PersistentUrlController.php
```diff
@@ -60,8 +73,13 @@ public function index(string $uuid): RedirectResponse {
       // \Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber.
       // More information you could find in this article:
       // https://www.lullabot.com/articles/early-rendering-a-lesson-in-debugging-drupal-8
-      if ($entity instanceof TranslatableInterface) {
-        return new RedirectResponse($entity->toUrl()->toString(), 302, ['PURL' => '1']);
+      if ($entity instanceof EntityInterface) {
+        // Not all entity types will need to be linked to their canonical URLs,
+        // so we dispatch an event to allow to modify the resulting URL.
+        $event = new PersistentUrlResolverEvent($entity);
```
upchuk, line 79: Never good idea to dispatch the same event in two different places. So let's make a service that is responsible for resolving the link of a given entity. In that, we default to canonical and we dispatch the event. And we use that service here and in the filter.

modules/oe_content_persistent/src/Event/PersistentUrlResolverEvent.php
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Event used to resolve the URL of an entity from its PURL.
+ */
+class PersistentUrlResolverEvent extends Event {
+
+  /**
+   * The name of the event.
+   */
+  const NAME = 'oe_content_persistent.event.persistent_url_resolver';
```
upchuk, line 19: `oe_content_persistent.event.entity_url_resolver`

modules/oe_content_persistent/src/Event/PersistentUrlResolverEvent.php
```diff
@@ -0,0 +1,75 @@
...
+   */
+  protected $url = NULL;
+
+  /**
+   * PersistentUrlResolverEvent constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The content entity.
+   */
+  public function __construct(EntityInterface $entity) {
```
upchuk, line 41: ContentEntityInterface no?

Do we support config entities? Adjust everywhere to typehint using ContentEntityInterface

modules/oe_content_persistent/src/Event/PersistentUrlResolverEvent.php
```diff
@@ -0,0 +1,75 @@
...
+    return $this->entity;
+  }
+
+  /**
+   * Gets the URL.
+   *
+   * @return \Drupal\Core\Url|null
+   *   The resolved URL.
+   */
+  public function getUrl(): ?Url {
```
upchuk, line 61: Let's not return NULL here. And have a method `hasUrl()` that when dispatching we can check.

modules/oe_content_persistent/src/Form/PurlSettingsForm.php
```diff
@@ -4,14 +4,47 @@
...
+  /**
+   * Constructs a PurlSettingsForm object.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The factory for configuration objects.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   */
+  public function __construct(ConfigFactoryInterface $config_factory, EntityTypeManagerInterface $entity_type_manager) {
+    $this->setConfigFactory($config_factory);
```
upchuk, line 34: Call the parent constructor here, don't attempt to replicate what it does,

modules/oe_content_persistent/src/Form/PurlSettingsForm.php
```diff
@@ -41,6 +74,13 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#description' => $this->t('The base URL to use for building persistent URLs'),
     ];
 
+    $form['supported_entity_types'] = [
+      '#type' => 'checkboxes',
+      '#options' => $this->getEntityTypeOptions(),
+      '#default_value' => $config->get('supported_entity_types'),
+      '#title' => $this->t('Supported entity types'),
```
upchuk, line 81: What happens if none are the entity types are supported? How does the whole thing behave? Can you check? because we might need to make this required for at least 1.

modules/oe_content_persistent/src/Form/PurlSettingsForm.php
```diff
@@ -50,9 +90,25 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+   * Returns an array of available entity types in the site.
+   *
+   * @return array
+   *   The available entity types keyed by ID.
+   */
+  protected function getEntityTypeOptions() : array {
+    $entity_type_definitions = $this->entityTypeManager->getDefinitions();
+    $entity_type_options = [];
+    foreach ($entity_type_definitions as $definition) {
+      $entity_type_options[$definition->id()] = $definition->getLabel();
```
upchuk, line 109: Do we support config entities as well?

modules/oe_content_persistent/tests/src/Functional/PersistentUrlControllerTest.php
```diff
@@ -89,6 +90,22 @@ public function testPersistentUrlController(): void {
     // Not valid uuid.
     $this->drupalGet('/content/' . $this->randomString());
     $this->assertResponse(404);
+
+    // Create a new node that will be send to the home page by the controller.
```
upchuk, line 94: be sent..

modules/oe_content_persistent/src/Form/PurlSettingsForm.php
imanoleguskiza: Indeed we will trigger some exceptions. I made the field required.

modules/oe_content_persistent/config/install/oe_content_persistent.settings.yml
imanoleguskiza: You are right, I added an update path.

modules/oe_content_persistent/oe_content_persistent.post_update.php
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+declare(strict_types = 1);
+
+/**
+ * Update PURL configuration.
+ */
+function oe_content_persistent_post_update_00001(): void {
+  $purl_config = \Drupal::configFactory()->getEditable('oe_content_persistent.settings');
+  $purl_config->set('supported_entity_types', ['node' => 'node']);
```
upchuk, line 15: We cannot assume node. The service parameters could have been overridden so we need to load the service params and update with what the site that uses the module has set.

modules/oe_content_persistent/src/Form/PurlSettingsForm.php
```diff
@@ -50,9 +91,27 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+   * Returns an array of available entity types in the site.
+   *
+   * @return array
+   *   The available entity types keyed by ID.
+   */
+  protected function getEntityTypeOptions() : array {
+    $entity_type_definitions = $this->entityTypeManager->getDefinitions();
+    $entity_type_options = [];
+    foreach ($entity_type_definitions as $definition) {
+      if ($definition->entityClassImplements('Drupal\Core\Entity\ContentEntityInterface')) {
```
upchuk, line 110: Arent the definitions ContentEntityType and ConfigEntityType? Just check instanceof

Screenshot 2020-07-31 11 59 13

PR #245, opened 2020-08-13
Update EPIC-Event with master

julien-:


julien-:
Useless to use EPIC-Event branch insofar as Event has been already released

PR #246, opened 2020-08-13
EWPP-125: Make 'Venue' and 'Organiser' fields optional.

julien-:
### Description

Event ; make 'Venue' and 'Organiser' fields optional.

### Change log

- Changed:
  - Venue field is not mandatory
    - widget "Inline entity form - Complex" is now used
  - Organiser field is not mandatory
    - removed useless Validation
  - Behat tests about required fieds

### DIFF
--- a/modules/oe_content_event/config/install/core.entity_form_display.node.oe_event.default.yml
+++ b/modules/oe_content_event/config/install/core.entity_form_display.node.oe_event.default.yml
@@ -403,13 +403,17 @@ content:
     settings:
       form_mode: default
       revision: true
-      label_singular: ''
-      label_plural: ''
-      override_labels: false
-      collapsible: false
+      override_labels: true
+      label_singular: venue
+      label_plural: venues
+      collapsible: true
+      allow_new: true
+      match_operator: CONTAINS
       collapsed: false
+      allow_existing: false
+      allow_duplicate: false
     third_party_settings: {  }
-    type: inline_entity_form_simple
+    type: inline_entity_form_complex
     region: content
   oe_event_website:
     weight: 12
--- a/modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_contact.yml
+++ b/modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_contact.yml
@@ -31,5 +31,5 @@ settings:
     sort:
       field: _none
     auto_create: false
-    auto_create_bundle: general
+    auto_create_bundle: oe_general
 field_type: entity_reference_revisions
--- a/modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_contact.yml
+++ b/modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_contact.yml
@@ -7,8 +7,8 @@ dependencies:
     - oe_content_entity_contact.oe_contact_type.oe_general
     - oe_content_entity_contact.oe_contact_type.oe_press
   module:
+    - composite_reference
     - entity_reference_revisions
-    - oe_content
 third_party_settings:
   composite_reference:
     composite: true
--- a/modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
+++ b/modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
@@ -7,7 +7,6 @@ namespace Drupal\oe_content_event\Plugin\Validation\Constraint;
 use Drupal\node\NodeInterface;
 use Symfony\Component\Validator\ConstraintValidator;
 use Symfony\Component\Validator\Constraint;
-use Symfony\Component\Validator\Violation\ConstraintViolationBuilderInterface;
 
 /**
  * Checks if the event fields are provided if required.
@@ -23,7 +22,6 @@ class EventFieldsRequiredValidator extends ConstraintValidator {
       return;
     }
 
-    $this->validateOrganiserGroupFields($constraint, $node);
     $this->validateRegistrationGroupFields($constraint, $node);
 
     $online_required_fields = [
@@ -73,42 +71,6 @@ class EventFieldsRequiredValidator extends ConstraintValidator {
             ->addViolation();
         }
       }
-    }
-  }
-
-  /**
-   * Validate organiser information consistency.
-   *
-   * An organiser can either be a custom string or a reference to a corporate
-   * vocabulary, depending from the value of `oe_event_organiser_is_internal`.
-   *
-   * This tests that, if one is set, the other is always not, depending
-   * whether the organiser is marked as internal or not.
-   *
-   * @param \Symfony\Component\Validator\Constraint $constraint
-   *   The constraint object.
-   * @param \Drupal\node\NodeInterface $node
-   *   The node object.
-   */
-  protected function validateOrganiserGroupFields(Constraint $constraint, NodeInterface $node) {
-    $violation = NULL;
-    if ($node->get('oe_event_organiser_internal')->isEmpty() && $node->get('oe_event_organiser_name')->isEmpty()) {
-      $violation = $this->context->buildViolation('You have to fill in at least one of the following fields: @internal or @organiser_name', [
-        '@internal' => $node->getFieldDefinition('oe_event_organiser_internal')->getLabel(),
-        '@organiser_name' => $node->getFieldDefinition('oe_event_organiser_name')->getLabel(),
-      ]);
-    }
-
-    if ($violation instanceof ConstraintViolationBuilderInterface) {
-      // Highlight empty 'Organiser name' field.
-      (clone $violation)
-        ->atPath('oe_event_organiser_name')
-        ->addViolation();
-
-      // Highlight empty 'Internal organiser' field.
-      $violation
-        ->atPath('oe_event_organiser_internal')
-        ->addViolation();
     }
   }
 
--- a/modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
+++ b/modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
@@ -7,7 +7,6 @@ namespace Drupal\oe_content_event\Plugin\Validation\Constraint;
 use Drupal\node\NodeInterface;
 use Symfony\Component\Validator\ConstraintValidator;
 use Symfony\Component\Validator\Constraint;
-use Symfony\Component\Validator\Violation\ConstraintViolationBuilderInterface;
 
 /**
  * Checks if the event fields are provided if required.
@@ -23,7 +22,6 @@ class EventFieldsRequiredValidator extends ConstraintValidator {
       return;
     }
 
-    $this->validateOrganiserGroupFields($constraint, $node);
     $this->validateRegistrationGroupFields($constraint, $node);
 
     $online_required_fields = [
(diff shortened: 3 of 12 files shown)


modules/oe_content_entity/modules/oe_content_entity_venue/src/Entity/Venue.php
```diff
@@ -69,4 +70,17 @@
...
-class Venue extends CorporateEntityBase implements VenueInterface {}
+class Venue extends CorporateEntityBase implements VenueInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
+    $fields = parent::baseFieldDefinitions($entity_type);
+
+    $fields['name']->setRequired(FALSE);
```
ademarco, line 81: We can't set the entity label not required.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
```diff
@@ -76,42 +74,6 @@ protected function validateGroupFieldsEmpty(array $fields, Constraint $constrain
...
-      (clone $violation)
-        ->atPath('oe_event_organiser_name')
-        ->addViolation();
-
-      // Highlight empty 'Internal organiser' field.
-      $violation
-        ->atPath('oe_event_organiser_internal')
-        ->addViolation();
-    }
-  }
```
ademarco, line 113: We cannot remove this, we need to change it, meaning: if they are both empty then we simple ignore the constrain, while if at least one of them is not, then we apply it.

tests/features/event.feature
```diff
@@ -281,12 +281,8 @@ Feature: Event content creation
     And I fill in "Content owner" with "Committee on Agriculture and Rural Development"
     And I fill in "Responsible department" with "Audit Board of the European Communities"
     And I fill in "Teaser" with "Event teaser"
-    And I press "Save"
-    Then I should see the following error messages:
-      | error messages                                                                                 |
-      | You have to fill in at least one of the following fields: Internal organiser or Organiser name |
     # Make sure that errors related to the Organiser fields are fixed.
-    When I check "Organiser is internal"
+    And I check "Organiser is internal"
```
ademarco, line 285: This was correct, it starts another assertion block.

modules/oe_content_entity/modules/oe_content_entity_venue/src/Entity/Venue.php
julien-: Fixed : I let the field required but waiting [this issue ](https://www.drupal.org/project/inline_entity_form/issues/3099844) is fixed what is your suggestion for our usage ?
- Change Form display Widget ?
- Propose a patch for the issue ?

At this moment I updated the widget **Inline entity form - Simple** with **Inline entity form - Complex**

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
julien-: The code did not correspond to the description of the function so I updated the description and the code the the new behavior

tests/features/event.feature
julien-: Fixed

src/Utility/ContentFormUtilities.php
```diff
@@ -20,32 +20,24 @@ class ContentFormUtilities {
...
-        [
-          ':input[name="' . $checkbox_field . '[value]"]' => ['checked' => FALSE],
-        ],
-      ],
-    ];
+  public static function toggleFieldsWithCheckbox(array &$form, string $checkbox_field, string $field1, string $field2, $is_required = TRUE): void {
+    $condition_checked = [':input[name="' . $checkbox_field . '[value]"]' => ['checked' => TRUE]];
+    $condition_unchecked = [':input[name="' . $checkbox_field . '[value]"]' => ['checked' => FALSE]];
+
+    // Field 1.
```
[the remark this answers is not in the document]
julien-: Fixed

src/Utility/ContentFormUtilities.php
```diff
@@ -20,32 +20,24 @@ class ContentFormUtilities {
...
+    $form[$field1]['#states']['visible'] = $condition_checked;
+    if ($is_required) {
+      $form[$field1]['#states']['required'] = $condition_checked;
+    }
+
+    // Field 2.
```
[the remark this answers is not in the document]
julien-: Fixed

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
julien-: Yesterday I implemented as you wanted :

```
protected function validateOrganiserGroupFields(Constraint $constraint, NodeInterface $node) {
    // Organiser is not required, so all subfields can be empty.
    if ($node->get('oe_event_organiser_internal')->isEmpty() && $node->get('oe_event_organiser_name')->isEmpty()) {
      return;
    }

    $violation = NULL;

    if ($node->get('oe_event_organiser_is_internal')->isEmpty() && !$node->get('oe_event_organiser_internal')->isEmpty()) {
      $violation = $this->context->buildViolation('When @is_internal field is not checked, @internal field have to be empty.', [
        '@is_internal' => $node->getFieldDefinition('oe_event_organiser_is_internal')->getLabel(),
        '@internal' => $node->getFieldDefinition('oe_event_organiser_internal')->getLabel(),
      ]);
    }
    elseif (!$node->get('oe_event_organiser_is_internal')->isEmpty() && !$node->get('oe_event_organiser_name')->isEmpty()) {
      $violation = $this->context->buildViolation('When @is_internal field is checked, @organiser_name field have to be empty.', [
        '@is_internal' => $node->getFieldDefinition('oe_event_organiser_is_internal')->getLabel(),
        '@organiser_name' => $node->getFieldDefinition('oe_event_organiser_name')->getLabel(),
      ]);
    }

    if ($violation instanceof ConstraintViolationBuilderInterface) {
      // Highlight empty 'Organiser name' field.
      (clone $violation)
        ->atPath('oe_event_organiser_name')
        ->addViolation();

      // Highlight empty 'Internal organiser' field.
      $violation
        ->atPath('oe_event_organiser_internal')
        ->addViolation();
    }
  }
```

But today I remove that because I think it is not needed insofar as nothing is required, so nothing have to be checked.
So drone is NOW green

Please @ademarco let me know if I am wrong. thank you

drishu: I think julien is right here, there is only 1 field shown (depending on the checkbox) so if its filled its filled, if its not its not (basically the essence of optional)

tests/features/event.feature
```diff
@@ -57,6 +57,8 @@ Feature: Event content creation
```
yenyasinn, line 57: Locally I have an error in this test because I have two medias "Euro with miniature figurines". I've fixed it locally by removing following code:
```
    # Create a "Media AV portal photo".
    And the following AV Portal photos:
      | url                                                         |
      | https://audiovisual.ec.europa.eu/en/photo/P-038924~2F00-15  |
      | https://audiovisual.ec.europa.eu/en/photo/P-039321~2F00-04  |

    # Create a "Media AV portal photo".
    And the following AV Portal photos:
      | url                                                         |
      | https://audiovisual.ec.europa.eu/en/photo/P-038924~2F00-15  |
```
as I see there is Drupal\Tests\oe_media\Behat\AvPortalContext where `media_avportal_mock` module is enabled. This module also provides "Euro with miniature figurines" media image too.

Could you please try to run this test locally. Do you face with issue?

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
```diff
@@ -91,21 +91,37 @@ protected function validateGroupFieldsEmpty(array $fields, Constraint $constrain
    *   The node object.
    */
   protected function validateOrganiserGroupFields(Constraint $constraint, NodeInterface $node) {
-    $violation = NULL;
     if ($node->get('oe_event_organiser_internal')->isEmpty() && $node->get('oe_event_organiser_name')->isEmpty()) {
-      $violation = $this->context->buildViolation('You have to fill in at least one of the following fields: @internal or @organiser_name', [
+      return;
```
nagyad, line 95: Why do you return without anything? You could simply remove the whole if statement.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
```diff
@@ -91,21 +91,37 @@ protected function validateGroupFieldsEmpty(array $fields, Constraint $constrain
...
+    $oe_event_organiser_is_internal_value = (bool) $node->get('oe_event_organiser_is_internal')->value;
+
+    if (!$oe_event_organiser_is_internal_value && !$node->get('oe_event_organiser_internal')->isEmpty()) {
+      // If checkbox "Organiser is internal" isn't checked,
+      // field "Internal organiser" has to be empty.
+      $violation = $this->context->buildViolation('When @is_internal field is not checked, @internal field have to be empty.', [
+        '@is_internal' => $node->getFieldDefinition('oe_event_organiser_is_internal')->getLabel(),
         '@internal' => $node->getFieldDefinition('oe_event_organiser_internal')->getLabel(),
+      ]);
+    }
```
nagyad, line 108: Why was this implemented? Was not part of the ticket and there are no such requirements. This we do automatically in oe_content_event_node_presave() and should not throw validation error.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
```diff
@@ -91,21 +91,37 @@ protected function validateGroupFieldsEmpty(array $fields, Constraint $constrain
...
+
+    if ($oe_event_organiser_is_internal_value && !$node->get('oe_event_organiser_name')->isEmpty()) {
+      // If checkbox "Organiser is internal" is checked,
+      // field "Organiser name" has to be empty.
+      $violation = $this->context->buildViolation('When @is_internal field is checked, @organiser_name field has to be empty.', [
+        '@is_internal' => $node->getFieldDefinition('oe_event_organiser_is_internal')->getLabel(),
```
nagyad, line 114: Same here, these should be removed.

nagyad (changes requested):

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
yenyasinn: Yes, it can be removed. Idea was to exit not to check next 3 conditions.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
ademarco: We need to make sure only one of the two is filled, depending on the value of the checkbox, but now that those fields are not mandatory anymore, empty values are also admissible. So the following are all valid scenarios:

- "Is internal" TRUE - "internal" EMPTY
- "Is internal" FALSE - "name" EMPTY
- etc.

But the following can never happen:

- "Is internal" TRUE - "name" NOT EMPTY

modules/oe_content_event/config/install/core.entity_form_display.node.oe_event.default.yml
```diff
@@ -403,13 +403,17 @@ content:
...
+      label_plural: venues
+      collapsible: true
+      allow_new: true
+      match_operator: CONTAINS
       collapsed: false
+      allow_existing: false
+      allow_duplicate: false
     third_party_settings: {  }
-    type: inline_entity_form_simple
+    type: inline_entity_form_complex
```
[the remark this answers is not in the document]
yenyasinn: As I understood we should use complex widget for any entity references

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
nagyad: Please see the ticket implementing that OPENEUROPA-2550 and it is written: 
image

yenyasinn: @ademarco please discuss with @nagyad correct approach.

modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_contact.yml
```diff
@@ -7,8 +7,8 @@ dependencies:
     - oe_content_entity_contact.oe_contact_type.oe_general
     - oe_content_entity_contact.oe_contact_type.oe_press
   module:
+    - composite_reference
     - entity_reference_revisions
-    - oe_content
```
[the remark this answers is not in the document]
yenyasinn: Drupal removed it.

modules/oe_content_event/config/install/core.entity_form_display.node.oe_event.default.yml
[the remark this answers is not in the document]
nagyad: It is not the case with Venue. Wasn't part of the ticket so please restore it as it was because that was. In general, you should not change existing code based on assumptions and without clear request in the ticket, because you will likely cause regression and we have no documentation over why the change has been implemented.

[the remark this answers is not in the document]
yenyasinn: @ademarco @drishu As I see there were discussion about it. Please clarify what widget we should to use here.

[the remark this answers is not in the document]
ademarco: Yeah let's not change this right now.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
ademarco: Ye, @nagyad is right. The case is the following: if a malicious user fills in both name and reference, this code will give an error, while PO doesn't want that, he only wants the case to be taken care silently, which we already do in the presave. The right thing to do here is to just remove the validation.

ademarco (approved):

nagyad (approved):

PR #272, opened 2020-09-11
EWPP-63: Theme "Call for tender" full view

yenyasinn:
### Description

TenderNodeWrapper has been added.

### DIFF
--- a/modules/oe_content_tender/src/TenderNodeWrapperInterface.php
+++ b/modules/oe_content_tender/src/TenderNodeWrapperInterface.php
@@ -75,10 +75,10 @@ interface TenderNodeWrapperInterface {
   /**
    * Get tender deadline date.
    *
-   * @return \Drupal\Core\Datetime\DrupalDateTime
+   * @return \Drupal\Core\Datetime\DrupalDateTime|null
    *   Deadline date as Drupal datetime object.
    */
-  public function getDeadlineDate(): DrupalDateTime;
+  public function getDeadlineDate(): ?DrupalDateTime;
 
   /**
    * Check whether the tender has opening date.
@@ -87,6 +87,14 @@ interface TenderNodeWrapperInterface {
    *   Whereas the tender has opening date.
    */
   public function hasOpeningDate(): bool;
+
+  /**
+   * Check whether the tender has deadline date.
+   *
+   * @return bool
+   *   Whereas the tender has deadline date.
+   */
+  public function hasDeadlineDate(): bool;
 
   /**
    * Gets status of the tender.
--- a/modules/oe_content_tender/src/TenderNodeWrapperInterface.php
+++ b/modules/oe_content_tender/src/TenderNodeWrapperInterface.php
@@ -4,6 +4,7 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_content_tender;
 
+use Drupal\Component\Render\MarkupInterface;
 use Drupal\Core\Datetime\DrupalDateTime;
 
 /**
@@ -14,7 +15,7 @@ interface TenderNodeWrapperInterface {
   /**
    * Tender isn't available. Is used when opening date isn't defined.
    */
-  const TENDER_STATUS_NOT_AVAILABLE = 'N/A';
+  const TENDER_STATUS_NOT_AVAILABLE = 'n/a';
 
   /**
    * Tender is upcoming.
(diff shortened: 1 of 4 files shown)


modules/oe_content_tender/tests/src/Kernel/TenderKernelTestBase.php
```diff
@@ -0,0 +1,72 @@
...
+   * @var array
+   */
+  public static $modules = [
+    'field',
+    'field_group',
+    'datetime_range',
+    'entity_reference_revisions',
+    'link',
+    'image',
+    'inline_entity_form',
```
drishu, line 26: there are some modules here which are not needed, like inline_entity_form, lets remove them one by one to see which are actually needed, the less is loaded the better

drishu (changes requested):

modules/oe_content_tender/src/TenderNodeWrapperInterface.php
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Interface for the tender entity wrapper.
+ */
+interface TenderNodeWrapperInterface {
+
+  /**
+   * Tender isn't available. Is used when opening date isn't defined.
+   */
+  const TENDER_STATUS_NOT_AVAILABLE = 'n/a';
```
ademarco, line 18:
```suggestion
  const TENDER_STATUS_NOT_AVAILABLE = 'not_available';
```
So we don't have `/` constant values.

modules/oe_content_tender/src/TenderNodeWrapperInterface.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function isClosed(): bool;
+
+  /**
+   * Check whether the tender status is not available.
+   *
+   * @return bool
+   *   Whereas the tender status is not available.
+   */
+  public function isNotAvailable(): bool;
```
ademarco, line 65: Given that the status field is optional (can't recall right now) let's change this into:
```suggestion
  public function hasStatus(): bool;
```

modules/oe_content_tender/src/TenderNodeWrapperInterface.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function hasDeadlineDate(): bool;
+
+  /**
+   * Gets status of the tender.
+   *
+   * @return string
+   *   Tender status.
+   */
+  public function getTenderStatus(): string;
```
ademarco, line 105:
```suggestion
  public function getStatus(): string;
```

ademarco (changes requested):

ademarco (approved):

drishu (approved):

modules/oe_content_tender/tests/src/Kernel/TenderNodeWrapperTest.php
```diff
@@ -0,0 +1,201 @@
...
+    TenderNodeWrapper::getInstance($entity);
+  }
+
+  /**
+   * Test exception thrown if an unsupported entity bundle is passed.
+   */
+  public function testUnsupportedEntityBundle(): void {
+    $this->expectException(\InvalidArgumentException::class);
+    $this->expectExceptionMessage("The current wrapper only accepts 'node' entities of type 'oe_tender'.");
+    $node_type = NodeType::create([
```
22Alexandra, line 39:
```suggestion
NodeType::create([
      'type' => 'test',
      'name' => 'Test bundle',
    ])->save();
```

22Alexandra: $node_type variable is not used.

yenyasinn: $node_type is used on the line 43.

22Alexandra: Indeed, but you can save after closing the bracket, no?

modules/oe_content_tender/tests/src/Kernel/TenderNodeWrapperTest.php
```diff
@@ -0,0 +1,201 @@
...
+      'type' => 'test',
+      'name' => 'Test bundle',
+    ]);
+    $node_type->save();
+    $entity = Node::create([
+      'type' => 'test',
+    ]);
```
22Alexandra, line 46: Also, this should be saved. Currently, is only created.

ademarco: This is actually ok, here we are testing unsupported bundle.

22Alexandra (approved):

PR #286, opened 2020-09-17
EWPP-270: Add "Details" field group to news content type.

yenyasinn:
### Description

New field oe_reference has been added to the News CT.


modules/oe_content_news/config/post_updates/00003_update_form_display/core.entity_form_display.node.oe_news.default.yml
```diff
@@ -0,0 +1,260 @@
...
+      format_type: details
+      region: content
+      format_settings:
+        id: ''
+        classes: ''
+        description: ''
+        open: true
+        required_fields: true
+      label: 'Alternative titles and teaser'
+    group_details:
```
drishu, line 42: I would not add this group here, such grouping is done on ewcms side (if I understood correctly). We should only add groups for fields that belong together like an event name and date fields, or venue name and location, so on.

modules/oe_content_news/oe_content_news.post_update.php
```diff
@@ -23,3 +25,28 @@ function oe_content_news_post_update_00001_update_field_labels(array &$sandbox):
...
+ */
+function oe_content_news_post_update_00002(): void {
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_news') . '/config/post_updates/00002_create_fields');
+  \Drupal::service('config.installer')->installOptionalConfig($storage);
+}
+
+/**
+ * Update news node form display.
+ */
+function oe_content_news_post_update_00003(): void {
```
drishu, line 40: I don't see the point in splitting this, the way I see it there should be 1 updated per epic per entity type, I know it was requested in 3119; @ademarco what are the guidelines here?

drishu (dismissed):

modules/oe_content_news/config/post_updates/00003_update_form_display/core.entity_form_display.node.oe_news.default.yml
yenyasinn: I did it as in Project CT - there are groups of fields.

modules/oe_content_news/oe_content_news.post_update.php
ademarco: it's ok to split it per "type of task", so to say, so that if one of these fail in production we know where it failed more precisely.

modules/oe_content_news/config/install/core.entity_form_display.node.oe_news.default.yml
```diff
@@ -7,17 +7,55 @@ dependencies:
...
     - field.field.node.oe_news.oe_subject
     - field.field.node.oe_news.oe_summary
     - field.field.node.oe_news.oe_teaser
     - node.type.oe_news
   module:
     - datetime
+    - field_group
     - link
     - maxlength
+    - path
```
ademarco, line 21: We should remove this dependency and related field from all default displays, I see it's also on the other content type. @yenyasinn  could you please create a follow up for this?

modules/oe_content_news/config/install/core.entity_form_display.node.oe_news.default.yml
```diff
@@ -30,51 +68,61 @@ content:
...
     region: content
     settings: {  }
     third_party_settings: {  }
+  langcode:
+    type: language_select
+    weight: 12
+    region: content
+    settings:
+      include_locked: true
+    third_party_settings: {  }
```
ademarco, line 86: I think also this shouldn't be exported in the default view mode, what happens when we enable this module on a non multilingual site? Let's check this in the follow up that we are  going to open above.

modules/oe_content_news/config/install/core.entity_form_display.node.oe_news.default.yml
```diff
@@ -150,25 +209,31 @@ content:
         maxlength_js_truncate_html: true
     type: text_textarea
     region: content
+  path:
+    type: path
+    weight: 14
+    region: content
+    settings: {  }
+    third_party_settings: {  }
```
ademarco, line 217: As said above, we should remove this in a follow up.

modules/oe_content_news/oe_content_news.info.yml
```diff
@@ -5,7 +5,9 @@ package: OpenEuropa
 type: module
 core: 8.x
 dependencies:
+  - field_group:field_group
```
ademarco, line 8: We have to enable this in a update hook.

modules/oe_content_news/oe_content_news.info.yml
```diff
@@ -5,7 +5,9 @@ package: OpenEuropa
...
   - oe_content:oe_content
+  - oe_content:oe_content_reference_code_field
```
ademarco, line 10: Also this one.

modules/oe_content_news/oe_content_news.post_update.php
```diff
@@ -23,3 +25,28 @@ function oe_content_news_post_update_00001_update_field_labels(array &$sandbox):
     $field_config->save();
   }
 }
+
+/**
+ * Create oe_reference field in the news content type.
+ */
+function oe_content_news_post_update_00002(): void {
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_news') . '/config/post_updates/00002_create_fields');
+  \Drupal::service('config.installer')->installOptionalConfig($storage);
```
ademarco, line 34: The field doesn't get created on my local.

ademarco (changes requested):

modules/oe_content_news/config/install/core.entity_form_display.node.oe_news.default.yml
yenyasinn: Follow up has been created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-291

modules/oe_content_news/oe_content_news.info.yml
yenyasinn: Done

modules/oe_content_news/oe_content_news.post_update.php
yenyasinn: How do you test it? Have you cleared cache after update?

imanoleguskiza (approved):

ademarco (approved):

PR #296, opened 2020-09-25
EWPP-307: Project content type fixes.

nagyad:


22Alexandra (approved):

modules/oe_content_project/config/install/field.field.node.oe_project.oe_project_locations.yml
```diff
@@ -12,7 +12,7 @@ entity_type: node
 bundle: oe_project
 label: 'Project locations'
 description: ''
-required: false
+required: true
```
sergepavle, line 15: Looks like it is fine to leave this field as required (declared in specification) even if we have an issue with the widget of location module.

sergepavle (approved):

ademarco (approved):

PR #313, opened 2020-10-06
EWPP-330: Call for Tenders closed status should not depend on the opening date.

22Alexandra:


nagyad (approved):

imanoleguskiza (approved):

ademarco (dismissed):
.

modules/oe_content_call_tenders/src/CallForTendersNodeWrapper.php
```diff
@@ -78,16 +78,16 @@ public function getStatus(): string {
     $now = $this->getNow();
 
     $status = static::STATUS_NOT_AVAILABLE;
-    if (!empty($opening_date)) {
+    if (isset($opening_date)) {
       if ($now < $opening_date) {
         $status = static::STATUS_UPCOMING;
       }
       elseif ($opening_date <= $now && $now < $closing_date) {
```
imanoleguskiza, line 85: We haven't checked if the closing_date is set here, so if its not we might get an error (although the field is required).

imanoleguskiza (changes requested):

modules/oe_content_call_tenders/src/CallForTendersNodeWrapper.php
22Alexandra: True

nagyad (approved):

imanoleguskiza (approved):

PR #317, opened 2020-10-09
EWPP-393: Remove IEF patch after release.

nagyad:


imanoleguskiza (dismissed):

composer.json
```diff
@@ -80,10 +80,6 @@
             "drupal/field_group": {
                 "https://www.drupal.org/project/field_group/issues/2787179#comment-13467953": "https://www.drupal.org/files/issues/2020-02-17/2787179-highlight-html5-validation-45.patch"
             },
-            "drupal/inline_entity_form": {
```
yenyasinn, line 83: These changes exist in 1.0-rc8 but doesn't exist in minor version that currently is 1.0-rc7.
Might be we also should update minimum version?
`"drupal/inline_entity_form": "~1.0-rc8"`

composer.json
```diff
@@ -44,6 +44,9 @@
         "phpunit/phpunit": "~6.0",
         "symfony/dom-crawler": "~3.4"
     },
+    "_readme": [
+        "We explicitly require drupal/inline_entity_form version to be higher than rc7 to avoid known issues with the module."
+    ],
```
brummbar, line 49: The comment is too generic, we require rc8 because we need the bug fix in it, more than "known issues". Let's improve the comment.

brummbar (changes requested):

imanoleguskiza (approved):

composer.json
```diff
@@ -44,6 +44,9 @@
         "phpunit/phpunit": "~6.0",
         "symfony/dom-crawler": "~3.4"
     },
+    "_readme": [
+        "We explicitly require drupal/inline_entity_form version to be higher than rc7 to have the bug fix for the table layout when opening multiple inline entities."
+    ],
```
brummbar, line 49:
```suggestion
    "_readme": [
        "We explicitly require drupal/inline_entity_form version rc8 or higher as it contains necessary bug fixes."
    ],
```

brummbar (changes requested):

composer.json
ademarco: Done, thanks!

ademarco (approved):

brummbar (approved):

PR #323, opened 2020-10-15
EWPP-405: Ensure referenced entities are not deleted upon removal.

nagyad:


22Alexandra:
IEF adds the flag "removed_reference" and sets a default to "optional" which lets the user decide if the referenced entity has to be removed or not. Adding upgrade paths in component after it has been released would change the default fallback of IEF to "keep" as we need it in EWCMS where we use editorial workflow.

I think the upgrade path should be only in EWCMS. 
@ademarco

ademarco (approved):

yenyasinn (approved):

ademarco (approved):

PR #329, opened 2020-10-21
EWPP-373: Ensure the browser widget works also inside IEF.

upchuk:


modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -282,16 +283,230 @@ public function testFeaturedMediaEntityBrowserWidget(): void {
...
+      ],
+      '1' => [
+        'target_id' => '2',
+        'caption' => 'Image 2 caption',
+      ],
+    ];
+    $actual_values = $node->get('featured_media_field')->getValue();
+    $this->assertEquals($expected_values, $actual_values);
+
+    // Edit the node to remove the current first item in the form (Image 2).
```
nagyad, line 462: Item 0 is the one you remove and that is Image 1.

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -282,16 +283,230 @@ public function testFeaturedMediaEntityBrowserWidget(): void {
...
+
+    $remove = $this->getSession()->getPage()->find('css', 'input[data-drupal-selector="edit-article-field-form-inline-entity-form-entities-0-form-featured-media-field-0-current-items-0-remove-button"]');
+    $remove->press();
+    $this->assertSession()->assertWaitOnAjaxRequest();
+
+    // Assert the image was removed from the field.
+    $this->assertSession()->pageTextNotContains('Image 1');
+    $this->getSession()->getPage()->pressButton('Save');
+
+    // Assert the remaining caption should not remain.
```
nagyad, line 477: This is confusing. You are asserting that the remaining caption can't be saved when the image is removed.

nagyad (changes requested):

nagyad (dismissed):

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
```diff
@@ -282,16 +283,230 @@ public function testFeaturedMediaEntityBrowserWidget(): void {
...
+    // Open the IEF form of the embedded article.
+    $this->getSession()->getPage()->pressButton('Edit');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+
+    $remove = $this->getSession()->getPage()->find('css', 'input[data-drupal-selector="edit-article-field-form-inline-entity-form-entities-0-form-featured-media-field-0-current-items-0-remove-button"]');
+    $remove->press();
+    $this->assertSession()->assertWaitOnAjaxRequest();
+
+    // Assert the image was removed from the field.
+    $this->assertSession()->pageTextNotContains('Image 1');
```
imanoleguskiza, line 474: Here we should double check that the selection button is still available to make sure the original issue does not reappear.

imanoleguskiza (changes requested):

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -262,6 +276,25 @@ protected function formElementItemsEntities(FieldItemListInterface $items, int $
...
+        $data = reset($data);
+        $value['target_id'] = explode(':', $data)[1];
+        $items->get($delta)->setValue($value);
+        $entity = $entity_storage->load($items[$delta]->target_id);
+        if (!empty($entity)) {
+          $entities[] = $entity;
+        }
+        return $entities;
+      }
+    }
```
brummbar, line 296: The original widget has a submit callback that sets into the form the value and then rebuild the form. That should be enough to have the values correctly populated. I did delete all this code, copied from the parent widget:
```
'#submit' => [[get_class($this), 'entityBrowserValueUpdated']],
'#limit_validation_errors' => [array_merge($element['#field_parents'], [$this->fieldDefinition->getName()])],
'#executes_submit_callback' => TRUE,
``` 
Then you need to override the method as we handle one value per field only. This way we get completely rid of user input mingling. Tests are green so I assume it's ok.
Also [fago says](https://www.drupal.org/project/entity_browser/issues/2851580#comment-13732789) that the above code seems not useful

brummbar (changes requested):

modules/oe_content_featured_media_field/tests/src/FunctionalJavascript/FeaturedMediaEntityBrowserWidgetTest.php
upchuk: Indeed. Added assertions. My bad. Not the test fails without the fix.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
```diff
@@ -262,6 +276,25 @@ protected function formElementItemsEntities(FieldItemListInterface $items, int $
...
+    $input_value = NestedArray::getValue($form_state->getUserInput(), $element_path, $input_exists);
+    $value_exists = NestedArray::keyExists($form_state->getValues(), $element_path);
+
+    if (!$value_exists && $input_exists && !is_array($input_value['target_id'])) {
+      $data = empty($input_value['target_id']) ? [] : explode(' ', trim($input_value['target_id']));
+      if ($data) {
+        $data = reset($data);
+        $value['target_id'] = explode(':', $data)[1];
+        $items->get($delta)->setValue($value);
+        $entity = $entity_storage->load($items[$delta]->target_id);
```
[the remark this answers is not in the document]
upchuk: No longer relevant.

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
upchuk: True. However, it seems I still need to check the input for the case in which an element is removed (and prevent defaults from $items). Could not find a better way. But indeed, using the #submit, we dont need to anymore load values here,

imanoleguskiza (dismissed):

modules/oe_content_featured_media_field/src/Plugin/Field/FieldWidget/FeaturedMediaEntityBrowserWidget.php
brummbar: True, it works for the default widget cause they have the same code around.

brummbar (changes requested):
It's all good but we miss test coverage for the removal bug which was fixed with the user input part.

brummbar (approved):

imanoleguskiza (approved):

PR #339, opened 2020-11-06
ISSUE: Not alter non-coporate node bundles.

julien-:
### Description

Restrict fields add to only Corporate node bundles.

### Change log

- Changed:
Restrict fields add to only Corporate node bundles.

### Commands

```sh
drush cr

```

### DIFF
--- a/oe_content.module
+++ b/oe_content.module
@@ -17,8 +17,17 @@ use Drupal\link\LinkItemInterface;
  */
 function oe_content_entity_bundle_field_info(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions) {
   $fields = [];
-  // Not alter non-coporate node bundles.
-  if ($entity_type->id() !== 'node' || strpos($bundle, 'oe_') !== 0) {
+  // Not alter non-corporate node bundles.
+  if ($entity_type->id() !== 'node' || !(in_array($bundle, array(
+      'oe_call_tenders',
+      'oe_event',
+      'oe_news',
+      'oe_organisation',
+      'oe_page',
+      'oe_policy',
+      'oe_project',
+      'oe_publication',
+  )))) {
     return $fields;
   }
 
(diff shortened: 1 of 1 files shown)


oe_content.module
```diff
@@ -13,12 +13,12 @@ use Drupal\Core\Field\FieldStorageDefinitionInterface;
...
- * Implements hook_entity_base_field_info().
+ * Implements hook_entity_bundle_field_info().
  */
-function oe_content_entity_base_field_info(EntityTypeInterface $entity_type) {
+function oe_content_entity_bundle_field_info(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions) {
   $fields = [];
-
-  if ($entity_type->id() !== 'node') {
+  // Not alter non-coporate node bundles.
+  if ($entity_type->id() !== 'node' || strpos($bundle, 'oe_') !== 0) {
```
ademarco, line 21: This means that, if one site decides to call its bundle `oe_something` then it will get these base fields. We should move this hook into each content type's sub-module and properly match that content type bundle's ID.

ademarco (changes requested):

oe_content.module
julien-: @ademarco 
Please take a look in my second suggestion.

brummbar:
The base fields are configurable and they can be hidden in the form UI easily. No need for changes in our side.

oe_content.module
```diff
@@ -13,12 +13,21 @@ use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\link\LinkItemInterface;
 
 /**
- * Implements hook_entity_base_field_info().
+ * Implements hook_entity_bundle_field_info().
  */
-function oe_content_entity_base_field_info(EntityTypeInterface $entity_type) {
+function oe_content_entity_bundle_field_info(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions) {
```
brummbar, line 18: This hook cannot be implemented in the same way as the hook_entity_base_field_info() one.

brummbar (changes requested):
The PR has no update hook, and the suggested change is major as it will impact the SQL table structures. So this cannot be done.

imanoleguskiza:
Closing pull request as it was labeled as invalid by sdt

PR #353, opened 2020-11-28
EWPP-445: Add teaser and subject common fields.

ademarco:


nagyad (dismissed):

modules/oe_content_call_proposals/config/install/core.entity_form_display.node.oe_call_proposals.default.yml
```diff
@@ -230,31 +243,44 @@ content:
...
+  oe_teaser:
+    type: text_textarea
+    weight: 16
+    region: content
+    settings:
+      rows: 5
+      placeholder: ''
+    third_party_settings:
+      maxlength:
+        maxlength_js: null
```
22Alexandra, line 255: We need to set the chars limit to 150.

ademarco: Yeah, let's do this, although we are not consistent on all content types with this.

22Alexandra (dismissed):

22Alexandra (approved):

upchuk (approved):

PR #365, opened 2020-12-09
EWPP-568: Add graph for SDG.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


sergepavle:
Not necessary for oe_content as it gonna be used in EWCMS only.

Issue #370, opened 2021-01-07
Adding items to a multiple entities field not working


miguelst:
I’m just starting to move our project to Open europa and came up with the following issue. I hope you can give me some assistant if I'm missing something.
Thanks in advance for your help.

**Steps to reproduce**
Install project https://github.com/ec-europa/regio-ewrc-d8-reference
Install oe_content module: `composer require "openeuropa/oe_content"`
Enable openeuropa/oe_content: `drush en oe_content oe_content_page oe_content_event`

**Expected behavior**
In the add event form
after clicking “add new venue”
a venue is added.

**Actual behavior**
The ajax spinner stay spining forever.
![Captura de pantalla 2020-12-16 a las 17 23 33](https://user-images.githubusercontent.com/24765416/103884970-a0c31280-50df-11eb-9369-4e7d654ba6ff.png)

The browser console returns the following error: Uncaught RangeError: Maximum call stack size exceeded. 
<img width="1122" alt="Captura de pantalla 2020-12-16 a las 17 22 31" src="https://user-images.githubusercontent.com/24765416/103885064-c05a3b00-50df-11eb-8b57-6b39db990440.png">

Is impossible to create the event due the venue is a required field.

**Remarks**
The same behavior happens in all the content types with multiple entities fields included the page content type.

miguelst:
Finally I found that the issue is solved when checking the option "Use the administration theme when editing or creating content" in the appearence page. I submitted a pull request to add it to the documentation.

ademarco:
The issue seems to be related to the usage of the OpenEuropa Theme as admin theme. Closing as it's unrelated to the oe_content module.

PR #375, opened 2021-01-15
Add usage information to avoid stack overflow issues

miguelst:
## NEMA-3013
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/NEMA-3013

### Description
Fixes issue described in https://github.com/openeuropa/oe_content/issues/370
Please could you add the proposed modification in the readme file to help others in the same situation?

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ If you want to use OpenEuropa Content, enable the module:
 ```bash
 drush en oe_content
 ```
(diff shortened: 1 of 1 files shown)


ademarco:
Hello @miguelst , the OpenEuropa theme is not supposed to support content editing, I think we should add such a disclaimer on the oe_theme project, not here.

PR #383, opened 2021-01-30
EWPP-727: Fix persistent caching of skos entities.

brummbar:


imanoleguskiza:
Closing, oe_content includes the lates rdf_skos version which includes the caching fix.

PR #387, opened 2021-02-16
EWPP-768: Add Description list field type.

22Alexandra:


nagyad (changes requested):
According to the title the ticket, this should have been a standalone module called description_list_field prepared a developed to be contributed to drupal.org with it's dedicated repository under OpenEuropa organisation.

PR #388, opened 2021-02-18
EWPP-684: Add jobs to a person content type

yenyasinn:


modules/oe_content_person/config/install/field.field.oe_person_job.default.oe_role_reference.yml
```diff
@@ -0,0 +1,34 @@
...
+default_value_callback: ''
+settings:
+  handler: 'default:skos_concept'
+  handler_settings:
+    target_bundles: null
+    sort:
+      field: _none
+    auto_create: false
+    concept_schemes:
+      - 'http://publications.europa.eu/resource/authority/corporate-body'
```
yenyasinn: It has to be http://publications.europa.eu/resource/authority/role. Waiting https://github.com/openeuropa/triple-store-dev/pull/40.

modules/oe_content_person/oe_content_person.info.yml
```diff
@@ -5,17 +5,19 @@ type: module
 core: 8.x
 
 dependencies:
-  - drupal:path
```
yenyasinn, line 8: It was moved to oe_content_organisation.

src/Utility/ContentFormUtilities.php
```diff
@@ -51,10 +51,6 @@ public static function toggleFieldsWithCheckbox(array &$form, string $checkbox_f
    *   The value that makes the dependent field visible.
    */
   public static function toggleFieldByValue(array &$form, string $toggle_field, string $field, string $value): void {
-    if (!isset($form[$toggle_field])) {
```
yenyasinn, line 54: It was removed to be used in hook_inline_entity_form_entity_form_alter() where elements from the parent form don't exist in the `$form` variable.

tests/features/person.feature
```diff
@@ -81,25 +81,31 @@ Feature: Person content creation
...
-    And I fill in "Media item" with "Contact image" in the "Person contact" region
-    And I fill in "Caption" with "Person contact caption" in the "Person contact" region
-    And I fill in "Press contacts" with "http://example.com/press_contacts" in the "Person contact" region
+    And I fill in "Media item" with "Contact image" in the "Person contacts" region
+    And I fill in "Caption" with "Person contact caption" in the "Person contacts" region
+    And I fill in "Press contacts" with "http://example.com/press_contacts" in the "Person contacts" region
+    # Jobs field.
+    And I press "Add new person job"
+    And I wait for AJAX to finish
+    And I fill in "first" person job role reference field with "Asia-Pacific Economic Cooperation"
```
yenyasinn: Role term from triple store has to be used. It is temporary solution just for testing.

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +44,77 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+    $entity_form['oe_role_name']['#attributes']['style'] = 'display:none';
+    $entity_form['oe_role_reference']['widget'][0]['target_id']['#required'] = TRUE;
+  }
+
+  // Set states based on "What type of person are you adding?" field.
+  ContentFormUtilities::toggleFieldByValue($entity_form, 'oe_person_type', 'oe_role_name', 'non_eu');
+  ContentFormUtilities::toggleFieldByValue($entity_form, 'oe_person_type', 'oe_role_reference', 'eu');
+  ContentFormUtilities::toggleFieldByValue($entity_form, 'oe_person_type', 'oe_acting', 'eu');
+
+  $entity_form['oe_role_name']['widget'][0]['value']['#states']['required'] = [
```
drishu: if we are going with this trend of converting state calls to helper functions then lets add this require by value also, requireFieldByValue(..

modules/oe_content_person/src/Entity/PersonJobInterface.php
```diff
@@ -0,0 +1,12 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_content_person\Entity;
+
+use Drupal\oe_content_sub_entity\Entity\SubEntityInterface;
+
+/**
+ * Represents a Person job entity.
+ */
+interface PersonJobInterface extends SubEntityInterface {}
```
drishu, line 12: I don't understand this empty interface, it doesn't bring multiple behaviours together, just looks like exploding class hierarchy to me, but if there is a reason do let me know

modules/oe_content_person/tests/src/Kernel/PersonJobEntityTest.php
```diff
@@ -0,0 +1,87 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_content_person\Kernel;
+
+use Drupal\KernelTests\Core\Entity\EntityKernelTestBase;
+use Drupal\Tests\rdf_entity\Traits\RdfDatabaseConnectionTrait;
+
+/**
+ * Tests document reference entity.
```
drishu:
```suggestion
 * Tests 'Person job' entities.
```

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +44,77 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+function oe_content_person_inline_entity_form_entity_form_alter(array &$entity_form, FormStateInterface &$form_state) {
+  if ($entity_form['#entity_type'] !== 'oe_person_job') {
+    return;
+  }
+
+  // Set initial state when form is loaded. If field
+  // "What type of person are you adding?" is not set then all fields will
+  // be hidden.
+  $type = NULL;
+  $user_input = $form_state->getUserInput();
```
drishu, line 84: I tested a lot here and 
```
  $entity_form['oe_role_name']['widget'][0]['value']['#states']['invisible'] = [
    ':input[name="oe_person_type"]' => [
      'value' => '',// or '_none' or NULL or 'empty' => true, etc
    ],
  ];
```
should work, but it does not, while checking for a specific value like non_eu works.

Seems to me like a drupal states api bug, worth a issue on d.org ?

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +44,77 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+ * Validation callback for a "oe_person_jobs" element.
+ *
+ * @param array $element
+ *   The element being processed.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The current state of the form.
+ * @param array $complete_form
+ *   The complete form structure.
+ */
+function oe_content_person_oe_person_jobs_validate(array &$element, FormStateInterface $form_state, array &$complete_form): void {
```
drishu: wasn't presave better as it would cover the programatic creation to, or do we need both this and presave for some UX usecase ?

drishu (changes requested):

modules/oe_content_person/config/install/field.field.oe_person_job.default.oe_role_reference.yml
drishu: we have a new tag 1.10, so changes can be made

modules/oe_content_person/oe_content_person.module
yenyasinn: Don't know what to say here.

modules/oe_content_person/oe_content_person.module
yenyasinn: We can't implement presave() until answering on https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-684?focusedCommentId=4557191&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4557191 because loss of data can happen.

Also we need this code to clean up hidden field. In case such field has incorrect value form won't be validated.

modules/oe_content_person/src/Entity/PersonJobInterface.php
yenyasinn: It is done with consistency with corporate entities. See modules/oe_content_entity/modules/oe_content_entity_venue/src/Entity/VenueInterface.php, modules/oe_content_entity/modules/oe_content_entity_organisation/src/Entity/OrganisationInterface.php, modules/oe_content_entity/modules/oe_content_entity_contact/src/Entity/ContactInterface.php. 

We use interfaces to identify type of entity because main entity class can be replaced.

modules/oe_content_person/oe_content_person.module
yenyasinn: Let's leave it as is. Field name here isn't simple string as for other fields but array `['oe_role_name']['widget'][0]['value']`

modules/oe_content_person/config/install/field.field.oe_person_job.default.oe_role_reference.yml
yenyasinn: Fixed

modules/oe_content_person/oe_content_person.module
yenyasinn: We can't use oe_content_person_oe_person_job_presave() here because main entity is saved after Job is saved. Therefore old type is available here only.

We can use oe_content_person_node_presave() and load all referenced jobs with cleaning theirs fields. But it means that on one Person update we will save Jobs twice. Let's use simple form validation for now.

drishu: Thinking from a user scenario point of view, if the person switches in time from eu to non_eu (or vice versa) then oe_content_person_node_presave would not be ok anyway.
Regarding validation, please test the scenario where the user click on the IEF create button for job, and then changes the person type, the expected behaviour needs to be discussed. For example he used to be non eu, now ie eu, and the 2 jobs need to reflect this.

modules/oe_content_person/oe_content_person.module
drishu: this is the issue https://www.drupal.org/project/drupal/issues/1091852, but it needs review and is a core issue, so we stick to the getUserInput workaround

modules/oe_content_person/src/Entity/PersonJobInterface.php
drishu: I don't see how a concrete implementation of an abstract class can just loose its parent, the scenario described sounds more like strategy pattern, but indeed, needs to be discussed within the team

modules/oe_content_person/oe_content_person.module
drishu: no problem, use ```NestedArray::setValue($entity_form, $parents,``` and send ```['oe_role_name', 'widget', 0, 'value', '#states', 'required']``` as $parent parameter

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +46,102 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+ * Implements hook_inline_entity_form_entity_form_alter().
+ */
+function oe_content_person_inline_entity_form_entity_form_alter(array &$entity_form, FormStateInterface &$form_state) {
+  if ($entity_form['#entity_type'] !== 'oe_person_job') {
+    return;
+  }
+
+  // Set initial state when form is loaded.
+  $type = NULL;
+  $user_input = $form_state->getUserInput();
```
upchuk, line 84: Can't we use getValues?

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +46,102 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+  if (isset($user_input['oe_person_type'])) {
+    $type = $user_input['oe_person_type'];
+  }
+  if ($type !== 'eu') {
+    $entity_form['oe_role_reference']['#attributes']['style'] = 'display:none';
```
upchuk, line 89: Let's apply the `.hidden` class on it instead of custom style.

tests/features/person.feature
```diff
@@ -202,3 +219,61 @@ Feature: Person content creation
     And I press "Save"
     # We assert that the extra characters are actually truncated from the end of the string.
     Then I should not see "The text to remove."
+
+  @javascript
```
upchuk, line 227: Wouldn;t this test be better as Kernel? Much faster IMO. Cause you can remove the reference also programatically in a new revision and assert the referenced entity was not deleted

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +46,102 @@ function oe_content_person_node_presave(EntityInterface $entity) {
     }
   }
 }
+
+/**
+ * Implements hook_inline_entity_form_entity_form_alter().
+ */
+function oe_content_person_inline_entity_form_entity_form_alter(array &$entity_form, FormStateInterface &$form_state) {
```
upchuk, line 77: Test coverage for this is lacking or quite inexistent (at least from what I can see).

* We don't assert the #states part (if the type is switched, that the job fields are also switched)
* We don't assert that when we open the jobs form, *only* the correct fields are shown
* We are only testing with 1 job reference...delta-related things might break and we don;'t catch it

If easier, do functional test and not Behat for this part...

upchuk (dismissed):
I left some comments, please check them out.

Anto should check also in detail as there are some things I dont know as well. 

As for the form, this needs to be thoroughly tested by Silvian once we merge to make sure all is fine and all the use cases are tried.

modules/oe_content_person/oe_content_person.module
```diff
@@ -42,3 +46,102 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+      $form_state->setValue(['oe_person_jobs', 'form', 'inline_entity_form', 'entities', $delta, 'form', 'oe_acting'], []);
+    }
+  }
+
+  // Enforce to update Role field on the existing jobs if person type has been
+  // changed.
+  foreach (Element::children($complete_form['oe_person_jobs']['widget']['entities']) as $key) {
+    $item = $complete_form['oe_person_jobs']['widget']['entities'][$key];
+    if (!empty($item['#entity']) && !_oe_content_person_validate_person_job_entity($item['#entity'], $type)) {
+      $form_state->setError($element, t('Role field is required in the job @label', ['@label' => $item['#entity']->label()]));
```
ademarco: We could maybe be more descriptive, so we explain the nature of the problem to the editor, something like:

```
The role "@label" is not compatible with the type of person currently selected. Please edit the related job entry and fix its role accordingly. 
```

ademarco (dismissed):

modules/oe_content_person/oe_content_person.module
yenyasinn: I have not investigated why but oe_person_type value is absent in form state values in `oe_content_person_inline_entity_form_entity_form_alter()`.

Screenshot 2021-02-25 at 11 33 49

So user input is used. Exact conditions `if ($type !== 'eu')`, `if ($type !== 'non_eu')` are used to prevent form from fake data.

modules/oe_content_person/oe_content_person.module
yenyasinn: `display:none` is a consistent way with Drupal core and will work with any themes. Would be better leave it so.

tests/features/person.feature
yenyasinn: Current implementation is made as in other CTs: consultation.feature, project.feature, event.feature.

Should we create a follow up and provide common test for these use cases?

modules/oe_content_person/oe_content_person.module
yenyasinn: Added todo to the function.
Follow up has been created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-800.

modules/oe_content_person/oe_content_person.module
yenyasinn: Message has been updated.

tests/features/person.feature
upchuk: I waas wrong here, we need to test the UI.

modules/oe_content_person/oe_content_person.module
yenyasinn: it was discussed with @ademarco to leave it as is.

modules/oe_content_person/tests/src/Kernel/PersonJobEntityTest.php
```diff
@@ -0,0 +1,96 @@
...
+    ]);
+    $person_job->save();
+    $this->assertEquals('Default', $person_job->label());
+
+    // Show role as a label if they are defined.
+    $person_job->set('oe_role_name', 'Role name label')->save();
+    $this->assertEquals('Role name label', $person_job->label());
+
+    $person_job->set('oe_role_reference', ['http://publications.europa.eu/resource/authority/corporate-body/APEC'])->save();
+    $this->assertEquals('Asia-Pacific Economic Cooperation', $person_job->label());
```
yenyasinn, line 92: Need to add test case with Acting role.

modules/oe_content_person/config/install/field.field.node.oe_person.oe_teaser.yml
```diff
@@ -12,7 +12,7 @@ entity_type: node
 bundle: oe_person
 label: Teaser
 description: 'A short overview of the information on this page. The teaser will be displayed in list views and search engine results, not on the page itself. Limited to 150 characters for SEO purposes.'
-required: true
+required: false
```
drishu, line 15: in https://webgate.ec.europa.eu/fpfis/wikis/display/EWPP/Common+Fields+-+fields+featured+in+all+Content+Types it is stated that teaser is mandatory, if this is not the case for person then this has to be mentioned in the ticket

modules/oe_content_person/oe_content_person.module
```diff
@@ -57,3 +71,93 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+    }
+  }
+}
+
+/**
+ * Validation of Person job entity based on the person type.
+ *
+ * @param \Drupal\oe_content_person\Entity\PersonJobInterface $entity
+ *   Person job entity
+ * @param string $type
```
drishu, line 160: Let's add param description

modules/oe_content_person/oe_content_person.module
```diff
@@ -57,3 +71,93 @@ function oe_content_person_node_presave(EntityInterface $entity) {
...
+/**
+ * Implements hook_inline_entity_form_entity_form_alter().
+ */
+function oe_content_person_inline_entity_form_entity_form_alter(array &$entity_form, FormStateInterface &$form_state) {
+  if ($entity_form['#entity_type'] !== 'oe_person_job') {
+    return;
+  }
+
+  // Set initial state when form is loaded.
+  // @todo: cover visible and required states by functional test.
```
drishu: done in PersonEntityTest.php, we can remove the todo

modules/oe_content_sub_entity/tests/modules/oe_content_sub_entity_test/src/Entity/SubEntityTypeTest.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_content_sub_entity_test\Entity;
+
+use Drupal\oe_content_sub_entity\Entity\SubEntityTypeBase;
+
+/**
+ * Defines the sub entity type text entity.
```
drishu:
```suggestion
 * Defines the sub entity type test entity.
```

modules/oe_content_person/tests/src/Kernel/PersonNodeWrapperTest.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_content_person\Kernel;
+
+use Drupal\oe_content_person\PersonNodeWrapper;
+
+/**
+ * Tests the Person content type.
```
drishu: doesn't this rather cover the PersonNodeWrapper class?

modules/oe_content_person/tests/src/Kernel/PersonNodeWrapperTest.php
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+    $this->installEntitySchema('oe_person_job');
+  }
+
+  /**
+   * Tests the getPersonJobLabels method of the PersonNodeWrapper class.
```
drishu, line 25: @covers ::getPersonJobLabels

modules/oe_content_person/tests/src/Kernel/PersonEntityTest.php
```diff
@@ -0,0 +1,160 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_content_person\Kernel;
+
+/**
+ * Tests the Person content type.
+ */
+class PersonEntityTest extends PersonEntityTestBase {
```
drishu, line 10: Isn't this test duplicating the behat one ?

modules/oe_content_person/tests/src/Kernel/PersonJobEntityTest.php
```diff
@@ -0,0 +1,99 @@
...
+    ]);
+
+    module_load_include('install', 'oe_content');
+    oe_content_install();
+  }
+
+  /**
+   * Tests label of Person job entity default bundle.
+   */
+  public function testDefaulPersonJobLabel() {
```
drishu, line 79: isn't this already covered by PersonNodeWrapperTest.php ? each one of the expected labels there basically covers this I think

drishu (changes requested):

modules/oe_content_person/oe_content_person.module
yenyasinn: Test has been added

modules/oe_content_person/tests/src/Kernel/PersonEntityTest.php
yenyasinn: Yes, it is. @upchuk suggests to get rid of Behat tests.

modules/oe_content_person/tests/src/Kernel/PersonJobEntityTest.php
yenyasinn: PersonNodeWrapperTest. covers getPersonJobLabels() method.
This test covers label() method. They are dependent but they are different and are used in different cases.

modules/oe_content_person/config/install/field.field.node.oe_person.oe_teaser.yml
yenyasinn: @nagyad asked to make this field optional.

drishu (dismissed):

nagyad (approved):

imanoleguskiza (approved):

PR #407, opened 2021-04-12
EWPP-970: Add char limit to first name and last name fields.

nagyad:


22Alexandra (dismissed):

sergepavle (dismissed):

modules/oe_content_person/oe_content_person.post_update.php
```diff
@@ -0,0 +1,33 @@
...
+declare(strict_types = 1);
+
+use Drupal\Core\Config\FileStorage;
+
+/**
+ * Add length restrictions for title display fields.
+ */
+function oe_content_person_post_update_20004(): void {
+  // Update form display to add maxlength to first and last names.
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_person') . '/config/post_updates/20001_person_title');
```
upchuk, line 17: 20004 vs 20001

upchuk (changes requested):

upchuk (approved):

modules/oe_content_person/oe_content_person.post_update.php
```diff
@@ -0,0 +1,33 @@
...
+use Drupal\Core\Config\FileStorage;
+
+/**
+ * Add length restrictions for title display fields.
+ */
+function oe_content_person_post_update_20001(): void {
+  // Update form display to add maxlength to first and last names.
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_person') . '/config/post_updates/20001_person_title');
+  $entity_type_manager = \Drupal::entityTypeManager();
+  $display_values = $storage->read('core.entity_form_display.node.oe_person.default');
```
sergepavle, line 19: We have to update not this config, but field.storage.

modules/oe_content_person/config/post_updates/20001_person_title/core.entity_form_display.node.oe_person.default.yml
```diff
@@ -0,0 +1,400 @@
+langcode: en
```
sergepavle, line 1: we need to update field.storage config

sergepavle (changes requested):

modules/oe_content_person/oe_content_person.post_update.php
imanoleguskiza: It was decided together with business that for first and last name we don't update the storage, but rather only add the restriction in the form. If you see the ticket description it was updated to reflect this.

sergepavle (approved):

modules/oe_content_person/oe_content_person.post_update.php
sergepavle: I got it

PR #409, opened 2021-04-15
EWPP-986: Adds overview field.

22Alexandra:


modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -18,3 +19,34 @@ function oe_content_organisation_post_update_00001(): void {
...
+ */
+function oe_content_organisation_post_update_00003(): void {
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_organisation') . '/config/post_updates/00003_create_fields');
+  \Drupal::service('config.installer')->installOptionalConfig($storage);
+}
+
+/**
+ * Update entity form display.
+ */
+function oe_content_organisation_post_update_00004(): void {
```
upchuk, line 41: Why do we have 3 update paths here? Let's use only 1

upchuk (changes requested):

modules/oe_content_organisation/oe_content_organisation.post_update.php
22Alexandra: We did the same for other content types when we had new dependencies and fields as suggested by @ademarco in case something goes wrong we know which update failed.

upchuk (approved):

sergepavle (approved):

PR #414, opened 2021-04-21
EWPP-988: Adds Leadership and organisation related fields.

22Alexandra:


modules/oe_content_organisation/config/install/core.entity_form_display.node.oe_organisation.default.yml
```diff
@@ -77,11 +79,29 @@ content:
...
+    weight: 27
+    settings:
+      match_operator: CONTAINS
+      match_limit: 10
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+    type: entity_reference_autocomplete
+    region: content
+  oe_organisation_persons:
```
imanoleguskiza: This is now added in the submodule so it shouldn't be here.

modules/oe_content_organisation/config/post_updates/00004_update_form_display/core.entity_form_display.node.oe_organisation.default.yml
```diff
@@ -77,11 +80,29 @@ content:
...
+    weight: 27
+    settings:
+      match_operator: CONTAINS
+      match_limit: 10
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+    type: entity_reference_autocomplete
+    region: content
+  oe_organisation_persons:
```
imanoleguskiza: Same as above, this is handled in the submodule.

modules/oe_content_organisation/oe_content_organisation.info.yml
```diff
@@ -14,6 +14,8 @@ dependencies:
   - inline_entity_form:inline_entity_form
   - oe_media:oe_media_avportal
   - description_list_field:description_list_field
+  - oe_media:oe_media
+  - oe_content:oe_content_organisation_person
```
imanoleguskiza: We need to remove this dependency, the whole point is that users can enable the submodule if they want.

modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -23,16 +23,22 @@ function oe_content_organisation_post_update_00001(): void {
...
- * Enable new dependency.
+ * Enable new dependencies.
  */
 function oe_content_organisation_post_update_00002(): void {
-  \Drupal::service('module_installer')->install(['description_list_field']);
+  \Drupal::service('module_installer')->install([
+    'description_list_field',
+    'oe_media',
+    'oe_content_person',
+    'oe_content_organisation_person',
```
imanoleguskiza: I think lets remove this from here and enable it on ewcms.

imanoleguskiza (changes requested):

modules/oe_content_organisation/modules/oe_content_organisation_person/oe_content_organisation_person.install
```diff
@@ -0,0 +1,58 @@
...
+function oe_content_organisation_person_install(): void {
+  // Add the persons entity reference field to the entity form display.
+  /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+  $form_display = EntityFormDisplay::load('node.oe_organisation.default');
+  if ($form_display) {
+    $oe_organisation_persons = [
+      'weight' => 29,
+      'settings' => [],
+      'third_party_settings' => [],
+      'type' => 'options_select',
```
imanoleguskiza: This should be an autocomplete I believe, all of our entity reference fields use autocomplete right?

imanoleguskiza (changes requested):

modules/oe_content_organisation/modules/oe_content_organisation_person/oe_content_organisation_person.install
22Alexandra: Indeed, I was using the autocomplete on ewcms but I'll update in here as well.

imanoleguskiza (dismissed):

ademarco (changes requested):
Please change the notation below into `1.0.x-dev`, without `~`:

image

modules/oe_content_organisation/config/install/core.entity_form_display.node.oe_organisation.default.yml
```diff
@@ -3,14 +3,16 @@ status: true
 dependencies:
   config:
     - field.field.node.oe_organisation.body
-    - field.field.node.oe_organisation.oe_organisation_overview
+    - field.field.node.oe_organisation.oe_organisation_org_chart
+    - field.field.node.oe_organisation.oe_org_staff_search_link
```
ademarco: Let's rename this field, if too long let's not sacrifice the prefix, but rather the rest, let's try:

```suggestion
    - field.field.node.oe_organisation.oe_organisation_staff_link
```

modules/oe_content_organisation/config/install/core.entity_form_display.node.oe_organisation.default.yml
```diff
@@ -3,14 +3,16 @@ status: true
 dependencies:
   config:
     - field.field.node.oe_organisation.body
-    - field.field.node.oe_organisation.oe_organisation_overview
+    - field.field.node.oe_organisation.oe_organisation_org_chart
```
ademarco: Let's simply call this `oe_organisation_chart`:
```suggestion
    - field.field.node.oe_organisation.oe_organisation_chart
```

modules/oe_content_organisation/config/install/field.storage.node.oe_organisation_org_chart.yml
```diff
@@ -0,0 +1,19 @@
...
+id: node.oe_organisation_org_chart
+field_name: oe_organisation_org_chart
+entity_type: node
+type: entity_reference
+settings:
+  target_type: media
+module: core
+locked: false
+cardinality: -1
+translatable: true
```
ademarco: Shouldn't this be not translatable, so we can reference the same media entity across all node's translations?

modules/oe_content_organisation/config/post_updates/00003_create_fields/field.storage.node.oe_organisation_org_chart.yml
```diff
@@ -0,0 +1,19 @@
...
+id: node.oe_organisation_org_chart
+field_name: oe_organisation_org_chart
+entity_type: node
+type: entity_reference
+settings:
+  target_type: media
+module: core
+locked: false
+cardinality: -1
+translatable: true
```
ademarco: Same as above, this should be false.

modules/oe_content_organisation/modules/oe_content_organisation_person/README.md
```diff
@@ -0,0 +1,3 @@
+# OpenEuropa Organisation Person Reference
+
+This module provides the Persons field which allows editors to reference Persons within the Organisation.
```
ademarco: Add the following:

> This field needs to be shipped in a dedicated module to avoid circular dependencies with the Person content type module `oe_content_person` as that module depends, in turn, on the `oe_content_organisation` module.

modules/oe_content_organisation/modules/oe_content_organisation_person/README.md
```diff
@@ -0,0 +1,3 @@
+# OpenEuropa Organisation Person Reference
```
ademarco: Let's rename the module machine name as follows:
```
oe_content_organisation_person_reference
```

So we are consistent with the other module in the same directory, with similar purposes:

image

modules/oe_content_organisation/modules/oe_content_organisation_person/config/install/field.storage.node.oe_organisation_persons.yml
```diff
@@ -0,0 +1,18 @@
...
+id: node.oe_organisation_persons
+field_name: oe_organisation_persons
+entity_type: node
+type: entity_reference
+settings:
+  target_type: node
+module: core
+locked: false
+cardinality: -1
+translatable: true
```
ademarco:
```suggestion
translatable: false
```

modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -23,22 +23,27 @@ function oe_content_organisation_post_update_00001(): void {
...
 /**
- * Enable new dependency.
+ * Enable new dependencies.
  */
 function oe_content_organisation_post_update_00002(): void {
-  \Drupal::service('module_installer')->install(['description_list_field']);
+  \Drupal::service('module_installer')->install([
+    'description_list_field',
+    'oe_media',
+    'oe_content_person',
```
ademarco: The module does not depend from `oe_content_person`, the optional `oe_content_organisation_person(_reference)` does.

modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -51,6 +56,11 @@ function oe_content_organisation_post_update_00004(): void {
       ->updateFromStorageRecord($form_display, $form_display_values);
     $updated_form_display->save();
   }
+
+  // Install submodule to allow referencing Person content types.
+  \Drupal::service('module_installer')->install([
+    'oe_content_organisation_person',
+  ]);
```
ademarco: Since the main module doesn't depend from `oe_content_organisation_person` (because it cannot, or else we get the circular dependency), we should remove this from here, or else and update state would different from and install state.

ademarco (changes requested):
Add to oe_content_organisation's README the following part:


	## Referencing persons from an organisation

	According to DG COMM's requirements, an organisation should be able to reference multiple persons. Unfortunately, since we already had the `oe_content_person` module depending on the `oe_content_organisation`, we cannot store an entity reference field (referencing persons) within the `oe_content_organisation` module.

	For this reason we have created the `oe_content_organisation_person_reference` sub-module.

	If you want to comply with DG COMM requirements, install the `oe_content_organisation_person_reference` module only after you have installed the `oe_content_organisation`, and export your site's configuration: this will avoid the circular dependency problem.

	If, on the other hand, you have a module that needs to depend on the `oe_content_organisation` module and it needs to provide a reference to persons, then make sure that your module depends only on `oe_content_organisation_person_reference`, and not on `oe_content_organisation`, or you will encounter the circular dependency issue when enabling your module.

modules/oe_content_organisation/oe_content_organisation.info.yml
```diff
@@ -28,16 +29,20 @@ config_devel:
...
     - field.storage.node.oe_organisation_acronym
     - field.storage.node.oe_organisation_contact
     - field.storage.node.oe_organisation_logo
     - field.storage.node.oe_organisation_eu_org
     - field.storage.node.oe_organisation_eu_org_type
     - field.storage.node.oe_organisation_non_eu_org_type
+    - field.storage.node.oe_organisation_chart
     - field.storage.node.oe_organisation_org_type
     - field.storage.node.oe_organisation_overview
+    - field.storage.node.oe_organisation_staff_search_link
```
ademarco:
```suggestion
    - field.storage.node.oe_organisation_staff_link
```

modules/oe_content_organisation/oe_content_organisation.info.yml
```diff
@@ -28,16 +29,20 @@ config_devel:
     - field.field.node.oe_organisation.oe_organisation_eu_org
     - field.field.node.oe_organisation.oe_organisation_eu_org_type
     - field.field.node.oe_organisation.oe_organisation_non_eu_org_type
+    - field.field.node.oe_organisation.oe_organisation_chart
     - field.field.node.oe_organisation.oe_organisation_org_type
     - field.field.node.oe_organisation.oe_organisation_overview
+    - field.field.node.oe_organisation.oe_organisation_staff_search_link
```
ademarco:
```suggestion
    - field.field.node.oe_organisation.oe_organisation_staff_link
```

modules/oe_content_person/oe_content_person.module
```diff
@@ -10,6 +10,7 @@ declare(strict_types = 1);
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element;
+use Drupal\field\Entity\FieldConfig;
```
ademarco: This is not used, we can remove it:

image

ademarco (changes requested):

modules/oe_content_organisation/modules/oe_content_organisation_person_reference/oe_content_organisation_person_reference.install
```diff
@@ -0,0 +1,63 @@
...
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
+
+/**
+ * Implements hook_install().
+ */
+function oe_content_organisation_person_reference_install(): void {
+  // Add the persons entity reference field to the entity form display.
+  /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+  $form_display = EntityFormDisplay::load('node.oe_organisation.default');
+  if ($form_display) {
```
ademarco:
```suggestion
  if ($form_display instanceof EntityFormDisplayInterface) {
```

ademarco (changes requested):

ademarco (approved):

imanoleguskiza (approved):

PR #419, opened 2021-04-22
EWPP-1049: Address field should have only Country as mandatory.

yenyasinn:


sergepavle (commented):
We have some tests which can be updated according to changes in address fields.

UPD: Some examples where we can left only country and don't fill in other address fields:
_features/project.feature_
Creation of a Project content through the UI.
But also possible to update tests:
- _features/person.feature_
- _features/call_proposals.feature_
- _features/event.feature_
- _features/news.feature_
- _features/organisation.feature_
- _features/publication.feature_

yenyasinn:
> We have some tests which can be updated according to changes in address fields.
> 
> UPD: Some examples where we can left only country and don't fill in other address fields:
> _features/project.feature_
> Creation of a Project content through the UI.
> But also possible to update tests:
> 
> * _features/person.feature_
> * _features/call_proposals.feature_
> * _features/event.feature_
> * _features/news.feature_
> * _features/organisation.feature_
> * _features/publication.feature_

We don't usually test required/optional fields in the Behat tests. In these tests we have to ensure that all fields can be filled. So I left them as is. I've updated only one case in Project CT when Organisation entity is added because it is used twice.

modules/oe_content_entity/modules/oe_content_entity_venue/config/install/core.entity_view_display.oe_venue.oe_default.default.yml
```diff
@@ -13,6 +13,24 @@ targetEntityType: oe_venue
 bundle: oe_default
 mode: default
 content:
+  created:
```
drishu, line 16: is this here after a config devel export? asking because I dont see the relation to the ticket

yenyasinn: Yes, it is default state of the configuration after fresh Drupal installation.

drishu (dismissed):

sergepavle (dismissed):

22Alexandra (dismissed):

22Alexandra (dismissed):

sergepavle (dismissed):

imanoleguskiza (approved):

upchuk (approved):

PR #420, opened 2021-04-22
EWPP-879: Change reference field value population approach.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/Behat/Content/Contact/GeneralContactContext.php
+++ b/tests/Behat/Content/Contact/GeneralContactContext.php
@@ -53,7 +53,7 @@ class GeneralContactContext extends RawDrupalContext {
       switch ($key) {
         // Set Media entity reference fields.
         case 'Image':
-          $fields = $this->getReferenceField($field_config, 'media', $value);
+          $fields = $this->getReferenceField($field_config, $value);
           $scope->addFields($fields)->removeField($key);
           break;
 
--- a/tests/Behat/Content/Contact/GeneralContactContext.php
+++ b/tests/Behat/Content/Contact/GeneralContactContext.php
@@ -5,7 +5,6 @@ declare(strict_types = 1);
 namespace Drupal\Tests\oe_content\Behat\Content\Contact;
 
 use Drupal\DrupalExtension\Context\RawDrupalContext;
-use Drupal\field\Entity\FieldConfig;
 use Drupal\Tests\oe_content\Behat\Hook\Scope\BeforeParseEntityFieldsScope;
 use Drupal\Tests\oe_content\Traits\EntityLoadingTrait;
 use Drupal\Tests\oe_content\Traits\EntityReferenceTrait;
@@ -46,14 +45,10 @@ class GeneralContactContext extends RawDrupalContext {
     ];
 
     foreach ($scope->getFields() as $key => $value) {
-      $field_config = NULL;
-      if (isset($mapping[$key])) {
-        $field_config = FieldConfig::loadByName($scope->getEntityType(), $scope->getBundle(), $mapping[$key]);
-      }
       switch ($key) {
         // Set Media entity reference fields.
         case 'Image':
-          $fields = $this->getReferenceField($field_config, $value);
+          $fields = $this->getReferenceField($scope->getEntityType(), $scope->getBundle(), $mapping[$key], $value);
           $scope->addFields($fields)->removeField($key);
           break;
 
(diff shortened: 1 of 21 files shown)


imanoleguskiza (dismissed):

tests/Behat/Content/Contact/GeneralContactContext.php
```diff
@@ -45,10 +46,14 @@ public function alterGeneralContactFields(BeforeParseEntityFieldsScope $scope):
     ];
 
     foreach ($scope->getFields() as $key => $value) {
+      $field_config = NULL;
```
imanoleguskiza: I wonder if instead of duplicating this piece of code everywhere it maybe should be part of the getreferencefield method. I leave this comment for the next reviewer to give their opinion

sergepavle: Yes indeed it can be also done inside the method, but in this case, anyway, we have to transfer information about entity_type/bundle/field_name to the method. I stayed for now with an approach for passing ready-to-use information in getReferenceField() for keeping the method simple as possible.
But it is discussable.

22Alexandra (dismissed):

tests/Behat/Content/Contact/GeneralContactContext.php
```diff
@@ -45,10 +46,14 @@ public function alterGeneralContactFields(BeforeParseEntityFieldsScope $scope):
...
+      if (isset($mapping[$key])) {
+        $field_config = FieldConfig::loadByName($scope->getEntityType(), $scope->getBundle(), $mapping[$key]);
```
ademarco: Let's move that into the method, I don't see any reason to replicate this every time.

ademarco (changes requested):

22Alexandra (dismissed):

imanoleguskiza (dismissed):

tests/Traits/EntityReferenceTrait.php
```diff
@@ -12,23 +14,29 @@ trait EntityReferenceTrait {
...
+    $field_config = FieldConfig::loadByName($entity_type, $bundle, $field_name);
+    /** @var \Drupal\Core\Entity\EntityReferenceSelection\SelectionInterface $handler */
+    $handler = \Drupal::service('plugin.manager.entity_reference_selection')->getSelectionHandler($field_config);
     // Transform titles to ids and maintain the comma separated format.
     $items = explode(',', $labels);
     $items = array_map('trim', $items);
     $ids = [];
     foreach ($items as $item) {
-      $ids[] = $this->loadEntityByLabel($entity_type, $item)->id();
+      $found_entities = $handler->getReferenceableEntities($item);
```
ademarco: This is not a good way of doing that. Check for example how the taxonomy term handler implements that:

image

It runs all sort of checks that we don't want to have when we are simply creating an entity in our tests. For example: you might want to reference a non published term, to then test that it appears when you publish it and refresh the page.

Using that method above would make this simply impossible.

The ticket was suggesting the right way of doing this: just get the target type / bundle from the handler configuration and use it in an entity query.

ademarco (changes requested):

tests/Traits/EntityReferenceRevisionTrait.php
```diff
@@ -12,26 +14,53 @@ trait EntityReferenceRevisionTrait {
...
     // Transform titles to ids and maintain the comma separated format.
     $items = explode(',', $labels);
     $items = array_map('trim', $items);
     $ids = [];
     $revision_ids = [];
     foreach ($items as $item) {
-      $entity = $this->loadEntityByLabel($entity_type, $item);
-      $ids[] = $entity->id();
-      $revision_ids[] = $entity->getRevisionId();
+      if ($label_key = $target_entity_type->getKey('label')) {
```
ademarco: According to this method's signature, we are requiring that all referenced entities will have labels, or else what would we pass in `$labels` to start with? Please restore this to it's original version.

tests/Traits/EntityReferenceTrait.php
```diff
@@ -12,23 +14,50 @@ trait EntityReferenceTrait {
...
+    elseif (!empty($configuration['target_bundles']) && is_array($configuration['target_bundles'])) {
+      $query->condition($target_entity_type->getKey('bundle'), $configuration['target_bundles'], 'IN');
+    }
     // Transform titles to ids and maintain the comma separated format.
     $items = explode(',', $labels);
     $items = array_map('trim', $items);
     $ids = [];
     foreach ($items as $item) {
-      $ids[] = $this->loadEntityByLabel($entity_type, $item)->id();
+      if ($label_key = $target_entity_type->getKey('label')) {
```
ademarco: According to this method's signature, we are requiring that all referenced entities will have labels, or else what would we pass in `$labels` to start with? Please restore this to it's original version.

ademarco (changes requested):

ademarco (approved):

nagyad (approved):

PR #429, opened 2021-05-12
Update composer.json

mike-michal:
this patch was committed and breaks instalation


imanoleguskiza:
Patch has been there for months and nothing is broken, no details about the issue are given.

PR #431, opened 2021-05-26
EWPP-911: Drupal 9 compatibility fixes.

yenyasinn:


composer.json
```diff
@@ -7,48 +7,50 @@
...
         "composer/installers": "~1.5",
-        "drupal/core-composer-scaffold": "^8.8",
+        "drupal/core-composer-scaffold": "^8.9 || ^9",
         "drupal/address": "~1.7",
         "drupal/composite_reference": "~1.0-alpha2",
         "drupal/config_devel": "~1.2",
-        "drupal/datetime_testing": "1.x-dev",
-        "drupal/description_list_field": "1.0.x-dev",
+        "drupal/datetime_testing": "dev-3217303-drupal-9-compatibility",
+        "openeuropa/description_list_field": "1.0.x-dev",
```
yenyasinn, line 24:
```suggestion
        "drupal/description_list_field": "1.0.x-dev",
```
when module will be released.

22Alexandra (changes requested):
The hook_install() implementations need to be updated since it receives a bool parameter.

yenyasinn:
> The hook_install() implementations need to be updated since it receives a bool parameter.

Fixed

composer.json
```diff
@@ -7,48 +7,50 @@
...
-        "openeuropa/oe_link_lists": "~0.5.0",
-        "openeuropa/oe_multilingual": "~1.5",
-        "openeuropa/oe_time_caching": "^1.0.0",
+        "openeuropa/code-review": "~1.6.0",
+        "openeuropa/drupal-core-require-dev": "^8.9.14 || ^9",
+        "openeuropa/oe_corporate_countries": "dev-EWPP-914",
+        "openeuropa/oe_corporate_site_info": "dev-EWPP-1137",
+        "openeuropa/oe_link_lists": "master",
+        "openeuropa/oe_multilingual": "master",
+        "openeuropa/oe_time_caching": "master",
```
22Alexandra: These components already include the Drupal 9 compatibility fixes in their latest release.

22Alexandra (changes requested):
There are still hook_install() implementations missing the fix or even the check that we're in the middle of a config sync, such as:

- `oe_content_organisation_person_reference_install`
- `oe_content_news_install`

composer.json
yenyasinn: "openeuropa/oe_time_caching" has to be used from master branch since new version isn't released yet.

22Alexandra (dismissed):

composer.json
```diff
@@ -7,48 +7,50 @@
...
+        "drupal/maxlength": "~1.0-rc1",
         "easyrdf/easyrdf": "1.0.0 as 0.9.2",
         "openeuropa/oe_media": "~1.12",
-        "openeuropa/rdf_skos": "~1.0"
+        "openeuropa/rdf_skos": "1.x-dev"
     },
     "require-dev": {
         "composer/installers": "~1.5",
-        "drupal/core-composer-scaffold": "^8.8",
+        "drupal/core-composer-scaffold": "^8.9 || ^9",
```
sergepavle:
```suggestion
        "drupal/core-composer-scaffold": "^8.9 || ^9.1",
```

composer.json
```diff
@@ -7,48 +7,50 @@
...
         "instaclick/php-webdriver": "^1.4.7",
         "nikic/php-parser": "^3.1.5",
         "openeuropa/behat-transformation-context": "~0.1",
-        "openeuropa/code-review": "~1.5.0",
-        "openeuropa/drupal-core-require-dev": "^8.9.14",
-        "openeuropa/oe_corporate_countries": "~2.0",
-        "openeuropa/oe_corporate_site_info": "~1.0",
-        "openeuropa/oe_link_lists": "~0.5.0",
+        "openeuropa/code-review": "~1.6.0",
+        "openeuropa/drupal-core-require-dev": "^8.9.14 || ^9",
```
sergepavle:
```suggestion
        "openeuropa/drupal-core-require-dev": "^8.9.14 || ^9.1",
```

sergepavle (changes requested):
In general changes now looks good. Left very minor comments but have found additional issue wich was discovered with last version of drupal 9 core:
image
Probably it is could be follow up issue but don't sure.

composer.json
yenyasinn: Since we have main dependency "drupal/core": "^8.9.14 || ^9.1", version 9.0.* of these packages won't be used in any case.

yenyasinn:
> but have found additional issue wich was discovered with last version of drupal 9 core:

Entity reference module is used in D8, so we can't remove usage of it now.

sergepavle (approved):

upchuk (approved):

PR #456, opened 2021-08-14
EWPP-1368: Maps sources field for behat scenarios.

22Alexandra:


tests/Behat/Content/Node/NewsContentContext.php
```diff
@@ -40,6 +40,7 @@ public function alterNewsFields(BeforeParseEntityFieldsScope $scope): void {
       'Reference' => 'oe_reference_code',
       'Location' => 'oe_news_location',
       'Body text' => 'body',
+      'Sources' => 'oe_news_sources',
```
yenyasinn, line 43: Do we need to add mapping for "Persons" and "Last updated on" fields for further usage (in third-party sites, for example)?

22Alexandra: Added the `Last updated date` field but since in oe_content we only provide the storage for the persons field and we can't control the field label it would not help for any third-party site.

nagyad (dismissed):

sergepavle (dismissed):

tests/Behat/Content/Node/NewsContentContext.php
yenyasinn: This Behat context is used to provide data for News CT so it will be helpful.

yenyasinn (dismissed):

ademarco (changes requested):
We don't enable `oe_content_departments_field` in an update hook, so the oe_departments field is not imported correctly. Provide a separate hook for that, before importing the configuration.

ademarco (approved):

yenyasinn (approved):

PR #458, opened 2021-08-18
EWPP-1410: Create author sub-entity type and related reference storage field.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

PR #459, opened 2021-08-19
EWPP-1414: Apply default content owner value to authors field.

sergepavle:
### Description
preceding PR: https://github.com/openeuropa/oe_content/pull/458

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ overriding the language.content_setting configuration.
 This module provides a sub-module (oe_content_entity) with a set of Corporate content entities to be used.
 Check out the [README](/modules/oe_content_entity/README.md) of the module.
 
-The module sets the "Authors" field value automatically when new content is created. The default "Authors" field value (reference to "Corporate body" sub-entity) can be set in the "Basic site settings" administrative page. This functionality is available only if [OpenEuropa Corporate Site Information](https://github.com/openeuropa/oe_corporate_site_info) is enabled.
+The module sets the "Content Owner" field value automatically when new content is created. The default "Content Owner" field value can be set in the "Basic site settings" administrative page. This functionality is available only if [OpenEuropa Corporate Site Information](https://github.com/openeuropa/oe_corporate_site_info) is enabled.
+
+Besides, "Content Owner" field, "Authors" field also receives default value (reference to "Corporate body" sub-entity) from settings shipped by [OpenEuropa Corporate Site Information](https://github.com/openeuropa/oe_corporate_site_info) optional module.
 
 This module requires sub-module 'oe_content_sub_entity' for supporting the "Authors" (which previously was declared also in the main module) field which is referencing the "Author" sub-entity type. It is done for supporting different types of 'authors' like "Corporate body", "Person (Local)", "Organisation (Local)", "Link".
 
--- a/README.md
+++ b/README.md
@@ -10,10 +10,6 @@ overriding the language.content_setting configuration.
 This module provides a sub-module (oe_content_entity) with a set of Corporate content entities to be used.
 Check out the [README](/modules/oe_content_entity/README.md) of the module.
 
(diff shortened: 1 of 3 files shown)


tests/FunctionalJavascript/DefaultContentOwnerTest.php
```diff
@@ -79,4 +79,56 @@ public function testContentOwnerValues() {
...
+    $this->drupalGet('/node/add/oe_page');
+
+    $assert_session = $this->assertSession();
+
+    // Assert that Corporate body author values are set
+    // when creating a new node.
+    $this->getSession()->getPage()->pressButton('ief-oe_authors-form-entity-edit-0');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $assert_session->fieldValueEquals('oe_authors[form][inline_entity_form][entities][0][form][oe_skos_reference][0][target_id]', 'Directorate-General for Agriculture and Rural Development (http://publications.europa.eu/resource/authority/corporate-body/AGRI)');
+    $assert_session->fieldValueEquals('oe_authors[form][inline_entity_form][entities][0][form][oe_skos_reference][1][target_id]', 'Directorate-General for Budget (http://publications.europa.eu/resource/authority/corporate-body/BUDG)');
```
nagyad: Let's test that we can save the node correctly and the values are there correctly.

nagyad (changes requested):

nagyad (approved):

README.md
```diff
@@ -10,7 +10,7 @@ overriding the language.content_setting configuration.
 This module provides a sub-module (oe_content_entity) with a set of Corporate content entities to be used.
 Check out the [README](/modules/oe_content_entity/README.md) of the module.
 
-The module sets the "Content Owner" field value automatically when new content is created. The default "Content Owner" field value can be set in the "Basic site settings" administrative page. This functionality is available only if [OpenEuropa Corporate Site Information](https://github.com/openeuropa/oe_corporate_site_info) is enabled.
+The module sets the "Authors" field value automatically when new content is created. The default "Authors" field value (reference to "Corporate body" sub-entity) can be set in the "Basic site settings" administrative page. This functionality is available only if [OpenEuropa Corporate Site Information](https://github.com/openeuropa/oe_corporate_site_info) is enabled.
```
imanoleguskiza: We shouldn't replace "Content owner" for "Authors", both are set automatically when creating nodes.

oe_content.module
```diff
@@ -137,5 +137,13 @@ function oe_content_node_prepare_form(EntityInterface $entity, $operation, FormS
 
   if (!empty($content_owners)) {
     $entity->set('oe_content_content_owner', array_values($content_owners));
+    // Set default corporate body entity for authors field.
```
imanoleguskiza: Earlier in the method we check whether the entity has the "oe_content_content_owner" field or not and we return early if it does. This means that hypothetically a content that has a "oe_author" field but not a "oe_content_content_owner" field would not get the default value.

imanoleguskiza (changes requested):

oe_content.module
```diff
@@ -137,5 +137,13 @@ function oe_content_node_prepare_form(EntityInterface $entity, $operation, FormS
 
   if (!empty($content_owners)) {
```
imanoleguskiza: I think this check is not needed since we already assert $site_information->hasDefaultContentOwners() before, but check it out just in case.

imanoleguskiza (changes requested):

README.md
sergepavle: My bad. Misunderstood requirements. Indeed, we shouldn't touch old content owner fields. fixed.

nagyad (approved):

imanoleguskiza (approved):

PR #463, opened 2021-09-03
[EWPP-1408] EWPP-1434: Add extra "Author" sub-entity bundles in a separate module.

sergepavle:
### Description
Preceding PR https://github.com/openeuropa/oe_content/pull/459
### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_content_person/src/EventSubscriber/SubEntityPersonJobSubscriber.php
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Event subscriber for handing entity labels for "Person job" entity bundles.
+ */
+class SubEntityPersonJobSubscriber extends SubEntitySubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function generateLabel(ContentEntityInterface $entity) {
+    switch ($entity->bundle()) {
```
yenyasinn:
```suggestion
```
We don't need to check $entity->bundle() because it is already checked in `protected function applies()`.

modules/oe_content_sub_entity/modules/oe_content_sub_entity_extra_authors/tests/src/Kernel/ExtraAuthorEntityTest.php
```diff
@@ -0,0 +1,206 @@
...
+      'oe_content_organisation',
+      'oe_content_sub_entity_extra_authors',
+    ]);
+  }
+
+  /**
+   * Tests label of extra author entities.
+   */
+  public function testLabel(): void {
+
```
yenyasinn:
```suggestion
```

modules/oe_content_sub_entity/modules/oe_content_sub_entity_extra_authors/tests/src/Kernel/ExtraAuthorEntityTest.php
```diff
@@ -0,0 +1,206 @@
...
+          }
+        }
+        elseif (!empty($data['field_values'])) {
+          $field_values = $data['field_values'];
+        }
+        $author->set($data['reference_field_name'], $field_values);
+        $author->save();
+        $this->assertEquals($data['expected_label'], $author->label());
+      }
+
```
yenyasinn:
```suggestion
```

modules/oe_content_sub_entity/src/Event/SubEntityEvent.php
```diff
@@ -0,0 +1,74 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_content_sub_entity\Event;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Event used to form sub-entity type labels individually for each bundle.
```
yenyasinn:
```suggestion
 * Event is used to form sub-entity type labels individually for each bundle.
```

modules/oe_content_sub_entity/src/Event/SubEntityEvent.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+class SubEntityEvent extends Event {
+
+  /**
+   * The name of the event.
+   */
+  const LABEL_FORMATION = 'oe_content.event.subentity_label_formation';
```
yenyasinn:
```suggestion
  const LABEL_FORMATION = 'oe_content_sub_entity.label_formation';
```
There is a pattern when event is described in the special final class to protect event, please check build/core/modules/block_content/src/BlockContentEvents.php, build/core/lib/Drupal/Core/Session/AccountEvents.php, build/core/modules/config_translation/src/Event/ConfigTranslationEvents.php.

modules/oe_content_sub_entity/src/Entity/SubEntityBase.php
```diff
@@ -28,7 +30,13 @@ abstract class SubEntityBase extends ContentEntityBase implements SubEntityInter
    */
   public function label() {
     // Return the entity's bundle label by default.
-    return $this->get('type')->entity->label();
+    $label = $this->get('type')->entity->label();
+
+    $event = new SubEntityEvent($this);
+    $event->setLabel($label);
```
yenyasinn:
```suggestion
```
Default label can be generated in the SubEntityEvent because we pass entity there and it will allow to dispatch event in another place without providing default label.

modules/oe_content_sub_entity/src/Event/SubEntityEvent.php
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * AuthorEntityEvent constructor.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The content entity.
+   */
+  public function __construct(ContentEntityInterface $entity) {
+    $this->entity = $entity;
+  }
```
yenyasinn:
```suggestion
    // Default label.
    $this->label = $entity->get('type')->entity->label()
  }
```

modules/oe_content_sub_entity/modules/oe_content_sub_entity_extra_authors/src/EventSubscriber/SubEntityExtraAuthorsSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+ * Event subscriber for handing entity labels for "Author" entity type bundles.
+ */
+class SubEntityExtraAuthorsSubscriber extends SubEntitySubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function generateLabel(ContentEntityInterface $entity): ?string {
+    $label = NULL;
+    switch ($entity->bundle()) {
```
yenyasinn:
```
case 'oe_link':
   ...
   break;

default:
  $label = $this->defaultLabel($entity);
```

modules/oe_content_sub_entity/src/Entity/SubEntityBase.php
```diff
@@ -28,7 +30,13 @@ abstract class SubEntityBase extends ContentEntityBase implements SubEntityInter
...
+    $this->eventDispatcher()->dispatch(SubEntityEvent::LABEL_FORMATION, $event);
+
+    return $event->getLabel();
```
yenyasinn: Let's cache label to avoid possible multiple invocation of events:
```
protected $label;

public function label() {
  if (empty($this->label) {
    $event = new SubEntityEvent($this);
    $this->eventDispatcher()->dispatch(SubEntityEvent::LABEL_FORMATION, $event);
    $this->label = $event->getLabel();
  }
  return $this->label;
}
```

sergepavle: Unfortunately, it will not work because many entity types depend on field values which can be changed.
I did checking `label()` methods implementation of many entity types and didn't find any usage of cache in any way.
Do you think it is worth to apply anyway caching inside label() method?

yenyasinn: Without caching event will be triggered every time and it will be calculate over and over again.

yenyasinn: But might be you are right - some dependencies can be changed. Might be we need to have a way to refresh the label?

sergepavle: As I can see currently we don't use some external data and in fact, we will work with already cached data (or already instanciated).

modules/oe_content_sub_entity/modules/oe_content_sub_entity_extra_authors/src/EventSubscriber/SubEntityExtraAuthorsSubscriber.php
```diff
@@ -0,0 +1,58 @@
...
+      case 'oe_person':
+      case 'oe_organisation':
+        $label = $this->defaultLabel($entity);
```
yenyasinn: We don't need 
```
      case 'oe_person':
      case 'oe_organisation':
        $label = $this->defaultLabel($entity);
```

anymore since  `case 'oe_person':, case 'oe_organisation':` are default cases

modules/oe_content_sub_entity/src/Entity/SubEntityBase.php
yenyasinn: I would suggest to avoid triggering event every time.

```
public function label($update = FALSE) {
  if (empty($this->label) || $update) {
    $event = new SubEntityEvent($this);
    $this->eventDispatcher()->dispatch(SubEntityEvent::LABEL_FORMATION, $event);
    $this->label = $event->getLabel();
  }
  return $this->label;
}
```

modules/oe_content_sub_entity/src/Entity/SubEntityBase.php
```diff
@@ -26,9 +28,13 @@ abstract class SubEntityBase extends ContentEntityBase implements SubEntityInter
   /**
    * {@inheritdoc}
    */
-  public function label() {
-    // Return the entity's bundle label by default.
-    return $this->get('type')->entity->label();
+  public function label(bool $update = FALSE) {
```
yenyasinn: $label is undefined in class. Let's define it `protected $label;`

modules/oe_content_sub_entity/src/Entity/SubEntityBase.php
```diff
@@ -26,9 +28,13 @@ abstract class SubEntityBase extends ContentEntityBase implements SubEntityInter
...
+    if (empty($this->label) || $update) {
+      $event = new SubEntityEvent($this);
+      $this->eventDispatcher()->dispatch(SubEntityEvent::LABEL_FORMATION, $event);
+      $this->label = $event->getLabel();
+    }
+    return $this->label;
```
yenyasinn, line 45: We also can add here 
```
  public function save() {
    $this->label = NULL;
    return parent::save();
  }
```
it allows us not to touch modules/oe_content_person/tests/src/Kernel/PersonJobEntityTest.php with `$person_job->label(TRUE);`

tests/features/event.feature
```diff
@@ -155,7 +155,7 @@ Feature: Event content creation
       | Facebook   |
       | Livestream |
 
-  @javascript @av_portal
+  @javascript @av_portal @ev
```
yenyasinn: Why do we need `@ev` tag?

tests/features/publication.feature
```diff
@@ -69,6 +69,43 @@ Feature: Publication content creation
...
+    And I select "Person" from "oe_authors[actions][bundle]"
+    And I press the "Add new author" button
+    And I fill in "Person (value 1)" with "Jane Doe"
+    And I press the "Add another item" button in the "Authors field" region
+    And I fill in "Person (value 2)" with "Jane 2 Doe 2"
+    And I press the "Create author" button
+    And I should see "Jane Doe, Jane 2 Doe 2"
+
+    And the following Organisation Content entity:
+      | Title             | Org 1 |
```
yenyasinn:
```suggestion
      | Title | Org 1 |
```

tests/features/publication.feature
```diff
@@ -69,6 +69,43 @@ Feature: Publication content creation
...
+    And the following Organisation Content entity:
+      | Title             | Org 2 |
```
yenyasinn:
```suggestion
      | Title | Org 2 |
```

modules/oe_content_sub_entity/src/Event/SubEntityEvent.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Drupal\oe_content_sub_entity\Event;
+
+use Drupal\oe_content_sub_entity\Entity\SubEntityInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Event is used to form sub-entity type labels individually for each bundle.
+ */
+final class SubEntityEvent extends Event {
```
yenyasinn: We need to implement two separate classes `final SubEntityEvents` and `SubEntityLabelInformationEvent extends Event`

modules/oe_content_sub_entity/src/Event/SubEntityEvent.php
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * The name of the event.
+   */
+  const LABEL_FORMATION = 'oe_content.event.label_formation';
```
yenyasinn:
```suggestion
  const LABEL_FORMATION = 'oe_content_sub_entity.label_formation';
```

modules/oe_content_sub_entity/src/SubEntitySubscriberBase.php
```diff
@@ -0,0 +1,104 @@
...
+   *   Labels separated by comma.
+   */
+  protected function getReferencedEntityLabels(ContentEntityInterface $entity): string {
+    // Load referenced entities.
+    $entities = $entity->referencedEntities();
+
+    $labels = [];
+    foreach ($entities as $entity) {
+      if ($entity instanceof ContentEntityInterface && $entity->getEntityType()->hasKey('label')) {
+        $labels[] = $entity->label();
```
yenyasinn: I assume that it doesn't provide translated labels that can be important for Author sub entity. Please see https://github.com/openeuropa/oe_content/pull/463/files#diff-9347041f0d1e9f4bf0f750dca978f2990c0589dbff59feb376ae15465da8ce88R25. I can't test it because this branch should be rebased since main EWPP-1410 is updated.

sergepavle: I don't sure if it should work according to requirements where the reference field to Person bundle is declared as not translatable. The branch is rebased.

yenyasinn: Relation to the Person is untranslatable, but Person node is translatable, so we need to show french label from Person node on the french page.
Screenshot 2021-09-22 at 18 52 47

This example from the form, but if we will need to show label on the page it will be untranslatable. For our current needs it might be not an issue but in future who knows...

So I would suggest
```
$entity = \Drupal::service('entity.repository')->getTranslationFromContext($entity);
$labels[] = $entity->label();
```

yenyasinn: It was a suggestion. We can't use `\Drupal::service('entity.repository')` here. Container injection for the event you had to implement yourself...

yenyasinn (dismissed):

nagyad (dismissed):

ademarco (dismissed):

nagyad (approved):

ademarco (approved):

PR #465, opened 2021-09-06
EWPP-1485: Add publication collection related fields and settings.

nagyad:


modules/oe_content_publication/config/install/field.field.node.oe_publication.oe_publication_collection.yml
```diff
@@ -0,0 +1,22 @@
...
+  config:
+    - field.storage.node.oe_publication_collection
+    - node.type.oe_publication
+id: node.oe_publication.oe_publication_collection
+field_name: oe_publication_collection
+entity_type: node
+bundle: oe_publication
+label: 'This is a publication collection'
+description: ''
+required: true
```
nagyad, line 13: This field need to have a value always.

tests/features/publication.feature
```diff
@@ -161,3 +168,35 @@ Feature: Publication content creation
...
+    And I click "Edit"
+    And I select the radio button "Yes"
+    And I press "Save"
+    Then I should see the error message "Publications field is required."
+    When I fill in "Publications" with "Publication demo page"
+    And I press "Save"
+    And I click "Edit"
+    And I select the radio button "No"
+    And I press "Save"
+    Then I should see the error message "Files field is required."
```
yenyasinn: Scenario is called "Publication collection fields are cleaned up according to the selected options." but it tests form validation only, although would be nice to test hook_entity_presave also.

modules/oe_content_publication/oe_content_publication.module
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Custom validation for the Publication node form.
+ *
+ * @param array $form
+ *   The form.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state.
+ */
+function _oe_content_publication_node_validate(array $form, FormStateInterface $form_state): void {
+  if ($form_state->getValue(['oe_publication_collection', 0, 'value']) == 1) {
```
yenyasinn:
```suggestion
  if ($form_state->getValue(['oe_publication_collection', 0, 'value']) === 1) {
```

modules/oe_content_publication/oe_content_publication.module
```diff
@@ -0,0 +1,87 @@
...
+ * Implements hook_form_BASE_FORM_ID_alter() for the node form.
+ */
+function oe_content_publication_form_node_form_alter(array &$form, FormStateInterface $form_state) {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $form_state->getBuildInfo()['callback_object']->getEntity();
+  if ($node->bundle() !== 'oe_publication') {
+    return;
+  }
+
+  $form['oe_documents']['#states'] = [
```
yenyasinn, line 31: Have you tried code
```
  $form['oe_documents']['widget'][0]['target_id']['#states'] = [
    'required' => [
      'input[name="oe_publication_collection"]' => ['value' => '0'],
    ],
  ];
  $form['oe_publication_publications']['widget'][0]['target_id']['#states'] = [
    'required' => [
      'input[name="oe_publication_collection"]' => ['value' => '1'],
    ],
  ];
```
It enables js required hint 
Screenshot 2021-09-09 at 15 25 40

since 
```
  $form['oe_documents']['#states'] = [
    'visible' => [
      'input[name="oe_publication_collection"]' => ['value' => '0'],
    ],
  $form['oe_publication_publications']['#states'] = [
    'required' => [
      'input[name="oe_publication_collection"]' => ['value' => '1'],
    ],
```
actually doesn't work.

Then we can avoid of using `_oe_content_publication_node_validate`

modules/oe_content_publication/oe_content_publication.module
nagyad: I this case it can be a string 1, so it's better to use double ==.

modules/oe_content_publication/oe_content_publication.module
nagyad: I theme there is a notice error with the admin theme using the method above [0]['target_id'], anyhow we use validation in other places as well because it works even without JS.

tests/features/publication.feature
nagyad: We can assert that the file is not rendered and the publication is not rendered.

modules/oe_content_publication/tests/src/Kernel/PublicationEntityTest.php
```diff
@@ -0,0 +1,163 @@
...
+      'status' => 1,
+    ]);
+    $node->save();
+
+    // Create a document for Publication.
+    $file = file_save_data(file_get_contents(drupal_get_path('module', 'oe_media') . '/tests/fixtures/sample.pdf'), "public://sample.pdf");
+    $file->setPermanent();
+    $file->save();
+
+    $media = \Drupal::service('entity_type.manager')->getStorage('media')->create([
```
yenyasinn:
```suggestion
    $media = $this->container->get('entity_type.manager')->getStorage('media')->create([
```
since it is Kernel test.

modules/oe_content_publication/tests/src/Kernel/PublicationEntityTest.php
```diff
@@ -0,0 +1,163 @@
...
+      'oe_media_file_type' => 'local',
+      'oe_media_file' => [
+        'target_id' => (int) $file->id(),
+      ],
+      'uid' => 0,
+      'status' => 1,
+    ]);
+    $media->save();
+
+    $node_storage = \Drupal::service('entity_type.manager')->getStorage('node');
```
yenyasinn:
```suggestion
    $node_storage = $this->container->get('entity_type.manager')->getStorage('node');
```
since it is Kernel test.

modules/oe_content_publication/tests/src/Kernel/PublicationEntityTest.php
```diff
@@ -0,0 +1,163 @@
...
+    $collection->save();
+
+    // Assert that the value of the publications reference is not saved because
+    // the publication is not a collection.
+    $this->assertTrue($collection->get('oe_publication_publications')->isEmpty());
+    $this->assertFalse($collection->get('oe_documents')->isEmpty());
+
+    // Now switch the publication to collection and assert the documents field.
+    $collection->set('oe_publication_collection', 1)->save();
+    $this->assertTrue($collection->get('oe_documents')->isEmpty());
```
yenyasinn, line 160: Would be good to save value in 'oe_publication_publications' and ensure it exists if 'oe_publication_collection' is 1.

modules/oe_content_publication/oe_content_publication.module
```diff
@@ -0,0 +1,102 @@
...
+
+  $form['oe_publication_publications']['#states'] = [
+    'visible' => [
+      'input[name="oe_publication_collection"]' => ['value' => '1'],
+    ],
+  ];
+
+  // We need to check if the widget array structure exists, it can vary if a
+  // different widget is being used.
+  if (isset($form['oe_publication_publications']['widget'][0]['target_id'])) {
```
yenyasinn, line 55: I assume that `isset($form['oe_publication_publications']['widget'][0]['target_id'])` can trigger warnings if such array is absent. Might be better to use 
```suggestion
  if (!empty($form['oe_publication_publications']['widget'][0]['target_id'])) {
```

nagyad: Did you know that empty() calls isset()? So I don't see how isset() is wrong here.

yenyasinn (approved):

modules/oe_content_publication/src/Plugin/Validation/Constraint/PublicationFieldsRequiredValidator.php
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($node, Constraint $constraint) {
+    /** @var \Drupal\node\NodeInterface $node */
+    if (!isset($node) || $node->getType() !== 'oe_publication') {
+      return;
+    }
+
+    $is_collection = (bool) $node->get('oe_publication_collection')->value;
```
22Alexandra, line 24: This is a new field added via hook_post_update on existing sites so we need to make sure the entity has the field otherwise:
Screenshot 2021-09-15 at 14 56 31

modules/oe_content_publication/config/install/field.field.node.oe_publication.oe_publication_collection.yml
22Alexandra: It should not be required. As requested in the ticket, this field should use a checkbox widget in the form display. So, instead of using `options_buttons`, `boolean_checkbox` widget type should be used so we don't force this field to be required.

modules/oe_content_publication/tests/src/FunctionalJavascript/PublicationEntityTest.php
```diff
@@ -0,0 +1,90 @@
...
+      'oe_publication_date' => [
+        'value' => '2020-04-15',
+      ],
+      'oe_subject' => 'http://data.europa.eu/uxp/1000',
+      'oe_author' => 'http://publications.europa.eu/resource/authority/corporate-body/AASM',
+      'oe_content_content_owner' => 'http://publications.europa.eu/resource/authority/corporate-body/COMMU',
+      'uid' => 0,
+      'status' => 1,
+    ]);
+    $collection->save();
```
22Alexandra, line 73: The files field should be required and filled in while you set `'oe_publication_collection' => 0,`. Maybe use the user to create the node since this is included in the Kernel test anyway.

22Alexandra (changes requested):

modules/oe_content_publication/src/Plugin/Validation/Constraint/PublicationFieldsRequiredValidator.php
22Alexandra: I was wrong, checked with @brummbar.

modules/oe_content_publication/tests/src/FunctionalJavascript/PublicationEntityTest.php
nagyad: The point was to create a node via api, to speed up, and assert that the File field is required when form is submitted and the same for the Publications field when Collection is set to TRUE.

22Alexandra (dismissed):

yenyasinn (dismissed):

yenyasinn (approved):

22Alexandra (approved):

PR #477, opened 2021-10-07
EWPP-1616: Adds common Subject field to Organisation CT.

22Alexandra:


modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -69,3 +69,28 @@ function oe_content_organisation_post_update_00005(): void {
...
+function oe_content_organisation_post_update_00006(): void {
+  // Create the subject field.
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_organisation') . '/config/post_updates/00006_subject_field');
+  $config_manager = \Drupal::service('config.manager');
+  $entity_type_manager = \Drupal::entityTypeManager();
+  $config_record = $storage->read('field.field.node.oe_organisation.oe_subject');
+  $entity_type = $config_manager->getEntityTypeIdByName('field.field.node.oe_organisation.oe_subject');
+  $entity_storage = $entity_type_manager->getStorage($entity_type);
+  $entity = $entity_storage->createFromStorageRecord($config_record);
+  $entity->save();
```
upchuk: I would first check that the field doesnt already exist. The field has been there for a long time and sites might have just added the field to their sites so this would crash those update paths. Let's do a quick check that it's there first.

upchuk (changes requested):

modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -69,3 +69,28 @@ function oe_content_organisation_post_update_00005(): void {
   $field_config->set('label', 'Contacts');
   $field_config->save();
 }
+
+/**
+ * Creates the subject field.
+ */
+function oe_content_organisation_post_update_00006(): void {
+  // Create the subject field.
+  $storage = new FileStorage(drupal_get_path('module', 'oe_content_organisation') . '/config/post_updates/00006_subject_field');
```
nagyad: We started to prefix the numbers with the module version like 20006, we should start doing it here too.

nagyad (changes requested):

modules/oe_content_organisation/oe_content_organisation.post_update.php
```diff
@@ -69,3 +69,28 @@ function oe_content_organisation_post_update_00005(): void {
...
+  $config_manager = \Drupal::service('config.manager');
+  $entity_type_manager = \Drupal::entityTypeManager();
+  $config_record = $storage->read('field.field.node.oe_organisation.oe_subject');
+  $entity_type = $config_manager->getEntityTypeIdByName('field.field.node.oe_organisation.oe_subject');
```
upchuk: Don't you know the entity type?

upchuk (changes requested):

nagyad (approved):

upchuk (approved):

PR #486, opened 2021-11-08
[EWPP-1681] EWPP-1435: Provide a service to migrate author v1 to v2.

nagyad:


modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/src/AuthorSkosUpdater.php
```diff
@@ -0,0 +1,98 @@
...
+      ->loadMultipleRevisions(array_keys($ids));
+
+    /** @var \Drupal\node\NodeInterface $revision */
+    foreach ($revisions as $revision) {
+      // If the field has no value we can skip the creation of new revision on
+      // the sub-entity and leave the 'oe_authors' field empty.
+      if ($revision->get('oe_author')->isEmpty()) {
+        continue;
+      }
+
```
yenyasinn, line 67: Might be add a condition to prevent second migration? In this case update could be run few times if needed.
```
if (!$revision->get('oe_authors')->isEmpty()) {
        continue;
      }
```

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/src/AuthorSkosUpdater.php
```diff
@@ -0,0 +1,98 @@
...
+      // Save the value over the same revision.
+      $revision->set('oe_authors', $author);
+      $revision->setNewRevision(FALSE);
+      $revision->setSyncing(TRUE);
+      // We don't want the automatic version system to kick in and set a version
+      // based on a wrong revision comparison.
+      // @see EntityVersionWorkflowManager::updateEntityVersion()
+      $revision->entity_version_no_update = TRUE;
+      $revision->save();
+    }
```
yenyasinn, line 91: I've found that for reverted revisions field `revision_default` is set as 0 in the `oe_author_revision` table, but has to be 1, since it is the active revision.

Screenshot 2021-11-09 at 12 47 54

As you see everywhere in another places this revision is set as active:
Screenshot 2021-11-09 at 12 50 17

Any thoughts? It might be useless for us since Author is a corporate entity that won't be used separately, but in any case please have a look https://www.drupal.org/node/2936349, might be we should copy state of revision also?

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/src/AuthorSkosUpdater.php
yenyasinn:
```
$is_default = $revision->isDefaultRevision();
```

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/src/AuthorSkosUpdater.php
```diff
@@ -0,0 +1,98 @@
...
+      // the sub-entity and leave the 'oe_authors' field empty.
+      if ($revision->get('oe_author')->isEmpty()) {
+        continue;
+      }
+
+      // Get the value from the revision and add it to the sub entity field.
+      $author_value = $revision->get('oe_author')->getValue();
+      $author->set('oe_skos_reference', $author_value);
+      if (!$author->isNew()) {
+        $author->setNewRevision(TRUE);
```
yenyasinn, line 77: `$author->isDefaultRevision($is_default );`

yenyasinn (dismissed):

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/src/AuthorSkosUpdater.php
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
+      $container->get('entity_type.manager')
+    );
+  }
```
sergepavle: Are we still need this method. I thought required service was injected through service already.

sergepavle (dismissed):

yenyasinn (dismissed):

upchuk (approved):

sergepavle (approved):

PR #487, opened 2021-11-08
[EWPP-1539] EWPP-1669: Add event collection functionality.

yenyasinn:


22Alexandra (changes requested):
The build has a test failing on both highest and lowest so it can't be random. The epic branch might need to be updated.

yenyasinn:
> The build has a test failing on both highest and lowest so it can't be random. The epic branch might need to be updated.

PR has been restarted and it's green now.

22Alexandra (approved):

nagyad (approved):

PR #490, opened 2021-11-16
EWPP-1620: Make author field optional.

nagyad:


yenyasinn (dismissed):

22Alexandra (dismissed):

modules/oe_content_publication/oe_content_publication.post_update.php
```diff
@@ -209,3 +209,12 @@ function oe_content_publication_post_update_20002(): void {
   $field_config->setRequired(FALSE);
   $field_config->save();
 }
+
+/**
+ * Set author as optional.
+ */
+function oe_content_publication_post_update_20003(): void {
```
upchuk: Any site that uses oe_content and relies on this field being required now is gonna get it optional?

upchuk (changes requested):

modules/oe_content_publication/oe_content_publication.post_update.php
nagyad: Only for publications.

22Alexandra (approved):

upchuk (approved):

PR #494, opened 2021-12-01
EWPP-1803: Add event extra fields.

22Alexandra:
### Description

Adds Event v2 extra fields.

### Change log

- Added: Online only boolean field, Link field to more media files, Long text field for mode media files description / more links.

### DIFF
--- a/modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_online_only.yml
+++ b/modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_online_only.yml
@@ -1,0 +1,22 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.node.oe_event_online_only
+    - node.type.oe_event
+id: node.oe_event.oe_event_online_only
+field_name: oe_event_online_only
+entity_type: node
+bundle: oe_event
+label: 'Online only'
+description: ''
+required: false
+translatable: false
+default_value:
+  -
+    value: 0
+default_value_callback: ''
+settings:
+  on_label: 'On'
+  off_label: 'Off'
+field_type: boolean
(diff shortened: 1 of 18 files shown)


imanoleguskiza (changes requested):
Should we maybe hide the venue field if the online only field is checked? I think its worth to confirm this

modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_online_only.yml
```diff
@@ -0,0 +1,22 @@
...
+label: 'Online only'
+description: ''
+required: false
+translatable: false
+default_value:
+  -
+    value: 0
+default_value_callback: ''
+settings:
+  on_label: 'On'
```
imanoleguskiza, line 20: The labels here probablye need to be "Yes" and "No"

modules/oe_content_event/config/post_updates/20002_update_form_display/core.entity_form_display.node.oe_event.default.yml
```diff
@@ -3,7 +3,9 @@ status: true
 dependencies:
```
imanoleguskiza, line 3: Not sure if we should update the form display of the node, sites using this component might have customized form displays and this will mess with their configuration.

imanoleguskiza (dismissed):

modules/oe_content_event/config/post_updates/20002_update_form_display/core.entity_form_display.node.oe_event.default.yml
imanoleguskiza: Policy was changed, rmeoving

modules/oe_content_event/config/install/field.field.node.oe_event.oe_event_online_only.yml
imanoleguskiza: Not needed as we use checkboxes

imanoleguskiza (dismissed):

nagyad (approved):

imanoleguskiza (approved):

PR #505, opened 2022-01-17
EWPP-1903: Fix the authors formatter.

nagyad:


nagyad:
With the commit above, the dedicated test will fail, because we should not see the change to the order of the values made on author entity due to revision reference.

imanoleguskiza (approved):

yenyasinn (approved):

PR #512, opened 2022-02-08
EWPP-1915: Update to PHP8

yenyasinn:


22Alexandra (changes requested):
Since there are no failures within this PR, please check https://drone.fpfis.eu/openeuropa/oe_theme/6838/82 because the entity form display config files come from oe_content and they need to be fixed in here.

composer.json
```diff
@@ -6,45 +6,50 @@
...
     "require": {
-        "php": ">=7.3",
-        "drupal/core": "^8.9.14 || ^9.1",
-        "drupal/linkit": "^5.0-beta11",
-        "drupal/maxlength": "^1.0@beta",
+        "php": ">=7.4",
+        "drupal/core": "^9.2",
+        "drupal/linkit": "^5.0-beta11 || ^6.0",
+        "drupal/maxlength": "^1.0@beta || ^2.0",
         "easyrdf/easyrdf": "1.0.0 as 0.9.2",
```
22Alexandra: This dependency is part of rdf_skos.

hernani: Can be removed.

22Alexandra:
> Since there are no failures within this PR, please check https://drone.fpfis.eu/openeuropa/oe_theme/6838/82 because the entity form display config files come from oe_content and they need to be fixed in here.

Nmv, I was wrong.

hernani (approved):

22Alexandra (approved):

PR #533, opened 2022-04-27
EWPP-2188: Correct xpath selector in tests for ckeditor button.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/src/Behat/WysiwygContext.php
+++ b/tests/src/Behat/WysiwygContext.php
@@ -54,7 +54,7 @@ class WysiwygContext extends RawDrupalContext {
    */
   public function iShouldSeeWysiwygButton($field, $button) {
     $wysiwyg = $this->getWysiwyg($field);
-    $button_elements = $this->getSession()->getDriver()->find($wysiwyg->getXpath() . '//a[@title="' . $button . '"]');
+    $button_elements = $this->getSession()->getDriver()->find($wysiwyg->getXpath() . '//a[starts-with(@title, "' . $button . '")]');
     if (empty($button_elements)) {
       throw new \Exception("Could not find the '$button' button.");
     }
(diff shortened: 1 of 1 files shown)


22Alexandra (approved):

tests/src/Behat/WysiwygContext.php
```diff
@@ -54,7 +54,7 @@ public function assertNoWysiwyg($label) {
    */
   public function iShouldSeeWysiwygButton($field, $button) {
     $wysiwyg = $this->getWysiwyg($field);
-    $button_elements = $this->getSession()->getDriver()->find($wysiwyg->getXpath() . '//a[@title="' . $button . '"]');
+    $button_elements = $this->getSession()->getDriver()->find($wysiwyg->getXpath() . '//a[starts-with(@title, "' . $button . '")]');
```
upchuk, line 57: I'm not sure this is a good idea. We lose capability with this of asserting the actual name of the button.
Also, the test that is failing is not in Behat so we should look into that and waiting for ckeditor to complete the name.

upchuk (changes requested):

tests/src/Behat/WysiwygContext.php
sergepavle: Screenshot from 2022-04-29 16-22-58
Unfortunately in local docker, I am still able to reproduce failure in behat tests. But indeed, it is not always reproducible. (did testing without 'start-with' with strict selector).

sergepavle: Reverted. I've found another solution.

sergepavle: The main issue is that within CKEditor we cannot guarantee same order of initialization of plugins. So there is few scenarios and impacts:

1. drupallink plugin
2. link plugin
In this case, Ctrl+K keystroke will be set to drupallink, and for Ctrl+K also for link plugin.
In this case, CKEditor will not set keystroke for drupallink plugin and title of drupallink button will be 'Link'

Other case:
1. link plugin
2. drupallink plugin
In this case, Ctrl+L keystroke will be set to link, and Ctrl+K also for the drupallink plugin.
In this case, CKEditor will set Ctrl+L for link plugin, Ctrl+K for drupallink plugin and title of drupallink button will be  'Link (Ctrl+K)'.


For fixing this conflict we may set keystroke forcefully by CKEditor config.

PR #537, opened 2022-06-02
EWPP-2286: Add Description field for Person content type.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_content_person/config/install/field.storage.node.oe_person_description.yml
```diff
@@ -0,0 +1,18 @@
+langcode: en
```
22Alexandra, line 1: Why not using the body field storage as we do for the rest of the CTs?

modules/oe_content_person/oe_content_person.post_update.php
```diff
@@ -72,3 +75,56 @@ function oe_content_person_post_update_20003(): void {
...
+  // View display configuration to update.
+  $view_display_values = $storage->read('core.entity_view_display.node.oe_person.default');
+  $view_display_values['_core']['default_config_hash'] = Crypt::hashBase64(serialize($view_display_values));
+  $view_display = EntityViewDisplay::load($view_display_values['id']);
+  if ($view_display) {
+    $updated_view_display = \Drupal::entityTypeManager()
+      ->getStorage($view_display->getEntityTypeId())
+      ->updateFromStorageRecord($view_display, $view_display_values);
+    $updated_view_display->save();
+  }
```
22Alexandra: We should update the view display only in oe_theme where we take care of the rendering.

22Alexandra (changes requested):
There is a test failing to be fixed.

modules/oe_content_person/config/install/field.storage.node.oe_person_description.yml
sergepavle: I would not reuse the body field. Initially, that field was used for the main content of content type. But description field in person content type is more used as additional content.

modules/oe_content_person/oe_content_person.post_update.php
sergepavle: removed.

modules/oe_content_person/config/install/field.storage.node.oe_person_description.yml
22Alexandra: Alright.

22Alexandra (dismissed):

nagyad (dismissed):

nagyad (approved):

22Alexandra (approved):

PR #543, opened 2022-06-30
EWPP-2171: Fix abbreviation of the month of 'May' for Irish.

22Alexandra:


translations/oe_content-ga.po
```diff
@@ -61,8 +61,8 @@ msgstr "Már"
 msgid "Apr"
 msgstr "Aib"
 
-msgid "Jun"
-msgstr "Mei"
+msgid "May"
```
yenyasinn, line 64: I suppose we shouldn't remove `msgid "Jun"` and just add msgid "May" additionally

yenyasinn (approved):

sergepavle (approved):

PR #550, opened 2022-09-23
EWPP-2607: Fix maxlength_js_enforce class not applied.

nagyad:
After maxlength release 2.0.0-rc1 the module doesn't use the html attribute of maxlength that enforced the limit on fields.
The module takes care of that by configuration. In order to ensure the same behaviour this configuration has to be enabled for text fields for the module to enforce and truncate extra strings above the limit.

### DIFF
--- a/oe_content.post_update.php
+++ b/oe_content.post_update.php
@@ -7,8 +7,10 @@
 
 declare(strict_types = 1);
 
+use Drupal\Core\Entity\Entity\EntityFormDisplay;
 use Drupal\Core\Utility\UpdateException;
 use Drupal\field\Entity\FieldConfig;
+use Drupal\node\Entity\NodeType;
 use Drupal\user\Entity\Role;
 
 /**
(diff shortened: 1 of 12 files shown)


oe_content.post_update.php
```diff
@@ -129,3 +130,25 @@ function oe_content_post_update_00004(): void {
...
+
+/**
+ * Ensure truncate maxlength setting is applied for textfields.
+ */
+function oe_content_post_update_00005(): void {
+  $displays = EntityFormDisplay::loadMultiple();
+  foreach ($displays as $display) {
+    $changed = FALSE;
+    foreach ($display->getComponents() as $field_name => $component) {
+      if ($component['type'] !== 'string_textfield' && empty($component['third_party_settings']['maxlength'])) {
```
22Alexandra: I suggest we update only the entity form displays of node entities.
If we update the `maxlength_js_enforce` value we should do this for `text_textarea` as well.
Shouldn't be an "or" condition instead of "and"? Since there might be text fields without maxlength set.

We need to make sure `type` is set.
warning

22Alexandra (changes requested):

22Alexandra (dismissed):

yenyasinn (dismissed):

22Alexandra (approved):

yenyasinn (approved):

PR #551, opened 2022-10-19
EWPP-2706: Allow descriptin field without featured image in events.

nagyad:


sergepavle (approved):

22Alexandra (changes requested):
This is not related to the scope of the ticket.

PR #552, opened 2022-10-25
EWPP-2714: Remove required fields constraint for Event Description field group.

22Alexandra:


modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequired.php
```diff
@@ -26,4 +26,18 @@ class EventFieldsRequired extends Constraint {
    */
   public $message = '@name field is required.';
 
+  /**
+   * The error message shown when featured media is provided.
+   *
+   * @var string
+   */
+  public $mediaNotEmpty = 'The %legend and %body fields are required when %media is provided.';
```
sergepavle: As I understand body field should not be required if a media field value is provided.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequired.php
```diff
@@ -26,4 +26,18 @@ class EventFieldsRequired extends Constraint {
...
+
+  /**
+   * The error message shown when featured media legend is filled in.
+   *
+   * @var string
+   */
+  public $legendNotEmpty = 'The %media and %body fields are required when %legend is specified.';
```
sergepavle: As I understand body field should not be required if a legend field value is provided.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequiredValidator.php
```diff
@@ -104,4 +99,42 @@ protected function validateRegistrationGroupFields(Constraint $constraint, NodeI
...
+        ->setParameters($parameters)
+        ->addViolation();
+      return;
+    }
+    if (!$is_legend_empty && $is_media_empty) {
+      $this->context->buildViolation($constraint->legendNotEmpty)
+        ->atPath('oe_event_featured_media')
+        ->setParameters($parameters)
+        ->addViolation();
+    }
```
sergepavle: As I understand we still could set the legend field value and media field without body field.

modules/oe_content_event/src/Plugin/Validation/Constraint/EventFieldsRequired.php
22Alexandra: Why? All three fields were required if any of them was filled in and editors wanted to be able to provide only the body value so that's how I changed the validation. No validation if only body is filled in but still make all three required if media or legend are provided.

22Alexandra: Discussed with Tamas to adapt the frontend and leave body field optional.

tests/features/event.feature
```diff
@@ -440,16 +440,21 @@ Feature: Event content creation
 
     # Make sure that validation of the Description fields group works as expected.
     When I click "Edit"
-    And I fill in "Full text" with "Full text paragraph"
+    When I fill in "Use existing media" with "Euro with miniature figurines" in the "Description" region
```
sergepavle:
```suggestion
    And I fill in "Use existing media" with "Euro with miniature figurines" in the "Description" region
```

sergepavle (dismissed):

yenyasinn (dismissed):

sergepavle (approved):

yenyasinn (approved):

PR #562, opened 2023-01-04
EWPP-0000: Fix persistent url test and magic getter returning false

nagyad:
Fix the test after the `https://ec.europa.eu` is redirecting to `https://commission.europa.eu/`. The test should use the new default url in order to make redirect request that doesn't redirect further.


brummbar:
There's actually no magic, the `empty` operation uses an `isset` before accessing a value. So for magic properties, the `__isset` method has to be implemented together with `__get`. The core change at https://www.drupal.org/project/drupal/issues/3298731 didn't add the __isset.

hernani (approved):

brummbar (approved):

PR #573, opened 2023-03-03
EWPP-3056: Introduce new decimal budget field for project content type.

nagyad:


modules/oe_content_project/config/install/field.field.node.oe_project.oe_project_eu_contrib.yml
```diff
@@ -0,0 +1,22 @@
...
+bundle: oe_project
+label: 'EU contribution'
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  min: null
+  max: null
```
sergepavle: Probably have to limit the max value, otherwise if in some case user adds a number like `1234567890123456789` we will not have any form error but fatal error (on trying to save node). It is reproducible on EWPP side.


`Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'oe_project_eu_budget_value' at row 1: INSERT INTO "node__oe_project_eu_budget" ("entity_id", "revision_id", "bundle", "delta", "langcode", "oe_project_eu_budget_value") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3, :db_insert_placeholder_4, :db_insert_placeholder_5); Array ( [:db_insert_placeholder_0] => 12 [:db_insert_placeholder_1] => 20 [:db_insert_placeholder_2] => oe_project [:db_insert_placeholder_3] => 0 [:db_insert_placeholder_4] => en [:db_insert_placeholder_5] => 1.2345678901235E+18 ) in Drupal\Core\Entity\Sql\SqlContentEntityStorage->saveToDedicatedTables() (line 1393 of /var/www/html/build/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).`

sergepavle:
Also there are erorr in tests:
`1) Drupal\Tests\oe_content_project\Functional\CopyBudgetFieldCommandTest::testUpdateProjectNodes
PHPUnit\Framework\Exception: Fatal error: During inheritance of IteratorAggregate: Uncaught Return type of Symfony\Component\Process\Process::getIterator($flags = 0) should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice`

sergepavle (dismissed):

modules/oe_content_project/README.md
```diff
@@ -48,3 +48,19 @@ In order to apply the patches above add the following to your project's `compose
 ```
 
 For more information check `cweagans/composer-patches` documentation [here](https://github.com/cweagans/composer-patches).
+
+## Deprecating Project budget fields and removing as of 4.x version
```
brummbar: This is not 100% true, the fields are there but named differently.

modules/oe_content_project/src/Commands/CopyBudgetValues.php
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * The node storage.
+   *
+   * @var \Drupal\node\NodeStorageInterface
+   */
+  protected $nodestorage;
+
+  /**
+   * UpdatePersonData class constructor.
```
brummbar: #doubt

modules/oe_content_project/src/Commands/CopyBudgetValues.php
```diff
@@ -0,0 +1,172 @@
...
+      $this->messenger->addError($this->t('There was a problem with the batch.'));
+      return;
+    }
+
+    $processed = count($results['processed']);
+    if ($processed === 0) {
+      $this->messenger->addStatus($this->t('No nodes have been processed.'));
+    }
+    else {
+      $this->messenger->addStatus($this->t('@items nodes with values have been processed: @ids.', [
```
brummbar: Are we sure we want to list all the node IDs that we updated? What if we have one thousand nodes?
Anyway fine by me.

modules/oe_content_project/src/Commands/CopyBudgetValues.php
```diff
@@ -0,0 +1,172 @@
...
+
+    $this->entityTypeManager = $entityTypeManager;
+    $this->messenger = $messenger;
+    $this->nodeStorage = $this->entityTypeManager->getStorage('node');
+  }
+
+  /**
+   * Triggers the field value copy.
+   *
+   * @command project-budget-copy-values:run
```
brummbar: We need to namespace commands with our project name, like oe-content:project-budget-copy-values

modules/oe_content_project/tests/src/Functional/CopyBudgetFieldCommandTest.php
```diff
@@ -0,0 +1,110 @@
...
+        // Assert the value is in decimal format in the new field.
+        $this->assertEquals(123456.00, $node->get('oe_project_eu_budget')->value);
+      }
+      if ($node->get('oe_project_budget_eu')->isEmpty()) {
+        $this->assertTrue($node->get('oe_project_eu_contrib')->isEmpty());
+      }
+      else {
+        // Assert the value is in decimal format in the new field.
+        $this->assertEquals(654321.00, $node->get('oe_project_eu_contrib')->value);
+      }
```
brummbar, line 106: We can leave it like this now, but this type of tests is never good. Better to have the different nodes "named", like
```
$project_nodes['empty budged'] = ....
```
and then run specific assertions of each scenario
```
$this->assertTrue($node->get('oe_project_eu_budget')->isEmpty())

File: modules/oe_content_project/tests/src/Functional/CopyBudgetFieldCommandTest.php [hunk 7]
```diff
@@ -0,0 +1,110 @@
...
+    $values['oe_project_budget_eu'] = NULL;
+    $project_node = $storage->create($values);
+    $project_node->save();
+    $project_nodes[] = $project_node;
+
+    $this->drush('project-budget-copy-values:run');
+    $storage->resetCache();
+
+    foreach ($project_nodes as $node) {
+      // Assert we have only one revision after copy.
```
brummbar: So, if I have a node with 3 revisions, how many will be there after?
Probably this is better
```suggestion
      // Assert that no new revisions have been created.
```

brummbar (changes requested):

brummbar (approved):

sergepavle (approved):

PR #580, opened 2023-03-24
EWPP-2814: Adapt the timeline field processor.

22Alexandra:


modules/oe_content_timeline_field/src/TimelineFieldProcessor.php
```diff
@@ -5,17 +5,17 @@
 namespace Drupal\oe_content_timeline_field;
 
 use Drupal\Core\Field\FieldItemListInterface;
-use Drupal\tmgmt_content\DefaultFieldProcessor;
+use Drupal\oe_translation\TranslationSourceFieldProcessor\DefaultFieldProcessor;
```
upchuk: You are using the same class here regardless of whether it exists or not..

upchuk (changes requested):

upchuk (approved):

brummbar (approved):

PR #588, opened 2023-06-08
EWPP-3305: Fix date translations and remove duplicates.

nagyad:


translations/oe_content-el.po
```diff
@@ -189,45 +189,3 @@ msgstr "Παρασκευή"
 
 msgid "Saturday"
 msgstr "Σάββατο"
-
```
22Alexandra, line 192: All long month names need to be updated for EL.

translations/oe_content-fi.po
```diff
@@ -189,45 +189,3 @@ msgstr "perjantai"
 
 msgid "Saturday"
 msgstr "lauantai"
-
```
22Alexandra, line 192: All long month names need to be updated for FI.

translations/oe_content-hr.po
```diff
@@ -189,45 +189,3 @@ msgstr "petak"
 
 msgid "Saturday"
 msgstr "subota"
-
```
22Alexandra, line 192: Please update the long month names.

translations/oe_content-da.po
```diff
@@ -3,231 +3,189 @@ msgstr ""
...
+msgstr "november"
 
 msgctxt "Long month name"
 msgid "December"
-msgstr "December"
+msgstr "december"
 
 msgid "Jan"
-msgstr "Jan."
+msgstr "jan."
```
22Alexandra: Why this dot? I might misunderstand the excel sheet but the dot is after the day, not after the month:
DA - short month name

I think we might need to remove the dot from every month name.

translations/oe_content-ro.po
```diff
@@ -189,45 +189,3 @@ msgstr "vineri"
 
 msgid "Saturday"
 msgstr "sâmbătă"
-
```
22Alexandra, line 192: For RO I don't see any dot after the short month name, I think we need to remove it (seems outdated).

22Alexandra (changes requested):

translations/oe_content-da.po
nagyad: But what if that is their way to make the abbreviation?

22Alexandra: They use %B for the long month name and %b for short month name. Seems good now, thanks.

22Alexandra (approved):

sergepavle (approved):

PR #592, opened 2023-08-07
ELA-665: OE Featured Media formatter with Image style.

julien-:
### Change log

- Added:
1. Possibility tu use the Drupal rendering using **oe_featured_media** field as an Image with image_style
2. You can use "UI Patterns Layouts" module to rennder the **oe_featured_media** field as an ImageValueObject


brummbar:
Code not for oe_content.

PR #608, opened 2024-01-12
EWPP-3846: Refactor WISYWIG trait and context for CKEditor5.

nagyad:


sergepavle (approved):

composer.json
```diff
@@ -11,15 +11,14 @@
         "drupal/core": "^10",
         "drupal/datetime_range_timezone": "^1.0.0-alpha3",
         "drupal/linkit": "^6.0.0-beta4",
-        "drupal/maxlength": "^2.0.1",
+        "drupal/maxlength": "dev-3.x",
```
imanoleguskiza: I think it might be better to point to a specific commit instead of having dev

tests/src/Behat/WysiwygContext.php
```diff
@@ -52,15 +52,16 @@ public function assertNoWysiwyg($label) {
...
    */
-  public function iShouldSeeWysiwygButton($field, $button) {
+  public function iShouldSeeWysiwygButton(string $field, string $button): void {
     $wysiwyg = $this->getWysiwyg($field);
-    $button_elements = $this->getSession()->getDriver()->find($wysiwyg->getXpath() . '//a[@title="' . $button . '"]');
-    if (empty($button_elements)) {
-      throw new \Exception("Could not find the '$button' button.");
-    }
-    if (count($button_elements) > 1) {
-      throw new \Exception("Multiple '$button' buttons found in the editor.");
```
imanoleguskiza, line 62: Why did we lose this assertion?

imanoleguskiza (dismissed):

composer.json
nagyad: It is temporary until we get the new release, in order to have dependencies handled correctly.

tests/src/Behat/WysiwygContext.php
nagyad: You can't have the same button multiple times in an editor. Config doesn't allow it.

sergepavle (dismissed):

sergepavle (approved):

22Alexandra (approved):
blind for @imanoleguskiza

PR #616, opened 2024-03-18
EWPP-3540: Delete oe_author entities referencing a node being deleted.

22Alexandra:


fmmribeiro (dismissed):

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/oe_content_sub_entity_author.module
```diff
@@ -48,3 +49,28 @@ function oe_content_sub_entity_author_node_prepare_form(EntityInterface $entity,
   ]);
   $entity->set('oe_authors', [$corporate_body]);
 }
+
+/**
+ * Implements hook_entity_delete().
+ *
+ * Delete author sub entities referencing a node being deleted.
+ */
+function oe_content_sub_entity_author_entity_delete(EntityInterface $entity) {
```
nagyad, line 58: We need to test this code. If I remove this hook we won't have any failures on the pipeline.

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/oe_content_sub_entity_author.module
```diff
@@ -48,3 +49,28 @@ function oe_content_sub_entity_author_node_prepare_form(EntityInterface $entity,
   ]);
   $entity->set('oe_authors', [$corporate_body]);
 }
+
+/**
+ * Implements hook_entity_delete().
+ *
+ * Delete author sub entities referencing a node being deleted.
```
nagyad, line 56: Delete author sub entities when the referenced node is being deleted. Author sub entity can reference Organisation or a Person node.

nagyad (changes requested):

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/oe_content_sub_entity_author.module
22Alexandra: Adapted test coverage.

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/oe_content_sub_entity_author.module
22Alexandra: Author entity has a field referencing nodes and we might add more node bundles later. I'd keep this comment as it is, doesn't change much. The if statement should be clear enough.

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/tests/src/Kernel/AuthorFieldEntityTest.php
```diff
@@ -206,6 +206,16 @@ public function testAuthorsField(): void {
     $this->assertEquals('Org1', $links[4]->a->__toString());
     $this->assertEquals('Org2', $links[5]->a->__toString());
 
+    // Load all existing author entities and assert there are 4 entities.
+    $authors = $this->entityTypeManager->getStorage('oe_author')->loadMultiple();
+    $this->assertCount(4, $authors);
+    // Delete the organisation nodes referenced by $author4 entity.
+    $org_node1->delete();
+    $org_node2->delete();
```
nagyad, line 215: These 2 are only organisation nodes, if I remove the person check from the if argument in the hook, this test will still pass.

nagyad (changes requested):

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/tests/src/Kernel/AuthorFieldEntityTest.php
22Alexandra: Both organisation nodes are being referenced by the `$authors4` entity. So the test asserts there are 4 author entity before removing the organisation nodes. After removing the organisation nodes there are only 3 author entity because the entity referencing them got deleted.

The person node is referenced by `$author3`, what is the relation?

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/tests/src/Kernel/AuthorFieldEntityTest.php
```diff
@@ -77,6 +78,7 @@ protected function setUp(): void {
       'oe_content',
       'oe_content_departments_field',
       'oe_content_person',
+      'oe_content_person',
```
nagyad: Person is already enabled. See the line before this line.

nagyad (dismissed):

fmmribeiro (dismissed):

upchuk (approved):

hernani (approved):

PR #622, opened 2024-06-06
EWPP-4288: Remove already migrated fields.

nagyad:
This must go into 4.x branch!


modules/oe_content_project/README.md
```diff
@@ -51,16 +51,10 @@ For more information check `cweagans/composer-patches` documentation [here](http
 
 ## Replacing Project budget fields and removing previous float type fields as of 4.x version
 
-As of 4.x the following float type fields will be removed:
+As of 4.x the following float type fields had been removed:
```
22Alexandra, line 54: So these fields were supposed to be removed in the next major considering the previous statement. Are we going to 4.x or is it ok to remove the fields in a minor version?

22Alexandra (changes requested):
Is `AuthorSkosUpdater` still needed?

modules/oe_content_project/README.md
nagyad: We will have to merge this into a new 4.x branch and we will go to a new major after this.

nagyad:
The `AuthorSkosUpdater` is needed because we made it as optional since we are not deleting or deprecating oe_author fields in this component, except EWPP where it is a project decision. Therefore, another site/project can decide anytime to move from oe_author to oe_authors sub-entities, which is a good reason to keep the updater service since we are maintaining both fields in this module.

modules/oe_content_project/oe_content_project.post_update.php
```diff
@@ -67,3 +68,18 @@ function oe_content_project_post_update_30001(): void {
...
+ * Remove old budget fields after 4.x.
+ */
+function oe_content_project_post_update_40001(): void {
+  $fields_to_remove = [
+    'oe_project_budget',
+    'oe_project_budget_eu',
+  ];
+  foreach ($fields_to_remove as $field_name) {
+    if ($field_storage = FieldStorageConfig::load("node.$field_name")) {
+      $field_storage->delete();
```
sergepavle, line 82: For safety better anyway to check if the field exists before trying to remove it.

sergepavle:
Are we planning still to keep `PersonNodeUpdater` service?

sergepavle (dismissed):

22Alexandra (dismissed):

sergepavle (approved):

hernani (approved):

PR #639, opened 2024-11-15
EWPP-4927: Person job Role plaintext field should always be visible but required only for non-EU.

22Alexandra:


modules/oe_content_person/tests/src/FunctionalJavascript/PersonEntityTest.php
```diff
@@ -136,18 +136,43 @@ public function testPersonForm() {
...
+    $this->getSession()->getPage()->pressButton('Save');
+    $this->assertSession()->pageTextContains('Person John Doe has been updated.');
+
+    // Edit and assert the values were kept.
+    $this->drupalGet('/node/1/edit');
+    $job_region = $this->getSession()->getPage()->find('css', '#edit-oe-person-jobs-wrapper');
+    $job_region->pressButton('Edit');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->assertEmpty($this->getSession()->getPage()->findField('oe_person_jobs[form][inline_entity_form][entities][0][form][oe_role_reference][0][target_id]')->getValue());
+    file_put_contents('form.html', $this->getSession()->getPage()->getOuterHtml());
```
intelektron: A leftover 

BTW, it's possible to output all HTML from all `->drupalGet()` calls, like in standard Drupal tests. 

First, set these env vars in docker-compose overrides: 
```shell
      BROWSERTEST_OUTPUT_DIRECTORY: /var/www/html/some_existing_dir
      BROWSERTEST_OUTPUT_BASE_URL: "http://localhost:8080/build"
```
And then run phpunit like this
```shell
docker compose exec web vendor/bin/phpunit --printer=\\Drupal\\Tests\\Listeners\\HtmlOutputPrinter
```
I found this one recently, it was turned off because it generates tons of HTML 😉.

modules/oe_content_person/oe_content_person.module
```diff
@@ -156,30 +145,55 @@ function _oe_content_person_oe_person_jobs_validate(array &$element, FormStateIn
   // changed.
   foreach (Element::children($complete_form['oe_person_jobs']['widget']['entities']) as $key) {
     $item = $complete_form['oe_person_jobs']['widget']['entities'][$key];
-    if (!empty($item['#entity']) && !_oe_content_person_validate_person_job_entity($item['#entity'], $type)) {
+    if (empty($item['#entity'])) {
+      continue;
+    }
+    if ($type === 'eu') {
+      if ($item['#entity']->get('oe_role_reference')->isEmpty() && $item['#entity']->get('oe_role_name')->isEmpty()) {
```
intelektron: Just a suggestion, but maybe it would be more readable to assign these isEmpty checks to a variable like `$is_role_reference_empty`, or something similar.

modules/oe_content_person/oe_content_person.module
```diff
@@ -156,30 +145,55 @@ function _oe_content_person_oe_person_jobs_validate(array &$element, FormStateIn
...
+  if (!is_array($person_jobs)) {
+    return;
   }
 
-  return TRUE;
+  foreach ($person_jobs as $key => $person_job_values) {
+    if ($key === 'inline_entity_form') {
+      continue;
+    }
+    if (empty($person_job_values['oe_role_name'][0]['value']) && !isset($person_job_values['oe_role_reference'][0]['target_id'])) {
```
intelektron: The same suggestion as above.

intelektron (changes requested):

modules/oe_content_person/tests/src/FunctionalJavascript/PersonEntityTest.php
22Alexandra: Thanks!

intelektron (approved):

modules/oe_content_person/oe_content_person.post_update.php
```diff
@@ -116,3 +116,15 @@ function oe_content_person_post_update_30001(): void {
     $updated_form_display->save();
   }
 }
+
+/**
+ * Update the oe_role_name field's label.
+ */
+function oe_content_person_post_update_30002(): void {
```
sergepavle, line 123: Looks like it is more EWCMS related requirement. Should we do that on component level? But change is very minor so suppose it is fine.

sergepavle (approved):

modules/oe_content_person/oe_content_person.post_update.php
22Alexandra: It's a corporate content type, I think DGCOMM can decide for components as well and if the field doesn't have the original set label, I don't change it.

PR #641, opened 2024-11-26
EWPP-4927: Add role type select to EU Person job.

22Alexandra:


22Alexandra:
Won't add a new select to the Person job entity.

PR #650, opened 2025-04-04
EWPP-5129: Change default event speaker label.

intelektron:
### Description

Change default event speaker bundle label. It is necessary in case when we add more bundles in the project.


sergepavle (approved):

22Alexandra (changes requested):
Why was this change needed?

intelektron:
@22Alexandra Otherwise, the editor has to choose between `Default` and `Europa person`, which is misleading. We talked it through with @nagyad 

image

22Alexandra (approved):

PR #673, opened 2025-10-13
EWPP-6061: Support Drupal Core 11.2.

hernani:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (approved):

sergepavle (approved):

modules/oe_content_sub_entity/modules/oe_content_sub_entity_author/tests/src/Kernel/AuthorSkosUpdaterTest.php
```diff
@@ -80,6 +80,11 @@ protected function setUp(): void {
    * Tests update of author values to sub-entities by the service.
    */
   public function testNodeUpdate(): void {
+    // Skip this test in 11.2 as we can't save
+    // old revisions anymore like we did.
+    if (version_compare(\Drupal::VERSION, '11.2.0', '>=')) {
+      $this->markTestSkipped();
```
piotrsmykaj, line 86: It would be nice to put a reference URL with an explanation of what is obsolete. Does it mean that this scenario should not be tested in Drupal 11.2+, or should this test be adapted in the future?

modules/oe_content_sub_entity/modules/oe_content_sub_entity_person/tests/src/Kernel/PersonNodeUpdaterTest.php
```diff
@@ -121,6 +121,11 @@ protected function setUp(): void {
    * Tests update of person values to sub-entities by the service.
    */
   public function testNodeUpdate(): void {
+    // Skip this test in 11.2 as we can't save
+    // old revisions anymore like we did.
+    if (version_compare(\Drupal::VERSION, '11.2.0', '>=')) {
+      $this->markTestSkipped();
```
piotrsmykaj, line 127: It would be nice to put a reference URL with an explanation of what is obsolete. Does it mean that this scenario should not be tested in Drupal 11.2+, or should this test be adapted in the future?

phpunit.xml.dist
piotrsmykaj, line 1: This is not an executable file, so it should have kept 644 permissions.

PR #680, opened 2026-03-11
EWPP-6514: Support D11.3.

nagyad:


modules/oe_content_timeline_field/tests/src/Kernel/TimelineWidgetTest.php
```diff
@@ -22,6 +22,7 @@ class TimelineWidgetTest extends EntityKernelTestBase {
    * {@inheritdoc}
    */
   protected static $modules = [
+    'oe_content',
```
donquixote, line 25: We ran into this in oe_whitelabel, in a functional test, where modules are installed with dependencies.
The reason is that oe_content_timeline_field does not declare a dependency to oe_content.
Should it?

Not having the dependency allows to install oe_content_timeline_field without oe_content, as long as we don't use the widget. If that is intended, great.

But, if indeed we want to avoid a dependency, it would be better if widgets and field types would work without the base module.

So base classes could be moved elsewhere.

The base module oe_content is quite heavy with all the base fields it introduces.

nagyad: Fixed, moved to oe_content_timeline.

intelektron (dismissed):

modules/oe_content_redirect_link_field/src/PathProcessor/PathProcessorRedirectLink.php
```diff
@@ -58,7 +58,12 @@ public function __construct(RedirectLinkResolverInterface $redirect_link_resolve
    */
   public function processOutbound($path, &$options = [], ?Request $request = NULL, ?BubbleableMetadata $bubbleable_metadata = NULL) {
     try {
-      $router = new Router(\Drupal::service('router.route_provider'), \Drupal::service('path.current'), \Drupal::service('url_generator'));
+      if (version_compare(\Drupal::VERSION, '11.2.99', '>')) {
+        $router = new Router(\Drupal::service('router.route_provider'), \Drupal::service('path.current'));
+      }
+      else {
+        $router = new Router(\Drupal::service('router.route_provider'), \Drupal::service('path.current'), \Drupal::service('url_generator'));
+      }
```
donquixote: Just fyi, core has a `DeprecationHelper::backwardsCompatibleCall()`.

It is advertised that "The below templates help code editors and PHPStan understand the return value of this function.". But tbh I don't know if that is true, and what exactly the impact would be here.

The method has some special support for core version strings ending with `-dev`. But this should not matter to us.

The current if/else is fine, actually looks a bit simpler than `::backwardsCompatibleCall()` :)

src/CollapsibleWidgetBase.php
donquixote, line 1: Just a note:
Modules that need this base class now also need to make sure that `oe_content_timeline_field` module is installed, even if they use this deprecated base class from `oe_content`.
If we want to avoid this, we could put an explicit `require_once` at the top of this file.

```php
require_once dirname(__DIR__) . '/modules/oe_content_timeline_field/src/CollapsibleWidgetBase.php`;
```
(This avoids any Drupal calls to determine module path, which should be avoided at file include time)

But, perhaps it is ok for other modules to just require that submodule. Depends if that module would give you any unwanted stuff.

donquixote: Another alternative could be to have a special namespace directory with base classes and utilities, that is loaded through composer autoload, and does not belong to any module. Or a dedicated module for utilities.

donquixote: Also `class_alias()` could be used.. but I can't think of a benefit right now. Current solution with `extends` should work fine.

phpstan.neon.dist
donquixote, line 1: Just fyi: If you want conditional ignores per Drupal version (or per PHP version), you can do `phpstan.neon.php` like here in oe_whitelabel: https://github.com/openeuropa/oe_whitelabel/blob/1.x/phpstan.neon.php

This becomes relevant if you run phpstan for different versions - which can be useful to identify if you are calling a method which only exists in some Drupal versions.

runner.yml.dist
```diff
@@ -39,6 +39,7 @@ drupal:
         'namespace' => 'Drupal\sparql_entity_storage\Driver\Database\sparql',
         'driver' => 'sparql',
       ];
+      $config['system.performance']['cache']['page']['max_age'] = 0;
```
donquixote: It is interesting to see this, are there problems with the cache?
It can make development easier, but could also obfuscate problems.
(I think this setting is not used when running tests, or is it?)

phpstan.neon.dist
intelektron: I am afraid that in this case it will not work. I tried it and PHPStan checks both branches of DeprecationHelper::backwardsCompatibleCall anyway.

modules/oe_content_redirect_link_field/src/PathProcessor/PathProcessorRedirectLink.php
intelektron: I've just changed it, and I'll do the same for `oe_theme`

runner.yml.dist
intelektron: I pushed a version without it, let's see. Localy I am having some random behat issues after removing this

intelektron: Well, it fails while running this Behat code:

```behat
When the "Document reference" sub-entity "Document reference to My Document 3" is updated as follows:
      | Published | No |
    And I am visiting the "Consultation demo page" content
    Then I should not see "sample.pdf"
```

Because most likely we don't add a cache tag of the sub-entity to the parent node. For other subentity types it happens in a custom field widget.

The question is - if this is a theoretical issue, or something that can happen in a real installation. Can we actually save subentities without touching their parents?

intelektron: In this case, I had to change `And I am visiting the "Consultation demo page" content` into a page reload.

22Alexandra (dismissed):

22Alexandra (dismissed):

intelektron (dismissed):

22Alexandra (approved):

intelektron (approved):

PR #687, opened 2026-07-20
EWPP-6988: Test against Drupal 11.4.

nagyad:


src/Form/EntityTypeForm.php
```diff
@@ -61,11 +61,28 @@ public function form(array $form, FormStateInterface $form_state) {
...
+    // has never been decorated the list is empty and the current class is the
+    // original. Detect the available API so this keeps working on Drupal 10
+    // and 11.3 too.
+    $entity_type_definition = $entity_type->getEntityType();
+    if (method_exists($entity_type_definition, 'getDecoratedClasses')) {
+      $decorated_classes = $entity_type_definition->getDecoratedClasses();
+      $original_class = $decorated_classes ? reset($decorated_classes) : $entity_type_definition->getClass();
+    }
+    else {
+      $original_class_method = 'getOriginalClass';
```
intelektron, line 77: Can't we just call `$entity_type_definition->getOriginalClass()`?

If it causes any problem with PHPStan, perhaps ignoring it in this line would be more readable.

nagyad: It will still break on the rest of the versions where there is nothing to ignore.

intelektron (approved):

22Alexandra (approved):

